PR #22, opened 2017-04-05
Rubocop

bess:


bess:
@mark-dce I added @no-reply's proposed Hydra project style, and cleaned up our code to match. The biggest change was alphabetizing and rationalizing our Gemfile, and I made that a separate commit. In some cases I disabled rules inline (Class has too many lines, and Block has too many lines). I didn't want to disable those rules entirely, because I want them active as we go forward, but for the specific violations already in place it doesn't seem worthwhile to refactor those right now.

mark-dce:
@no-reply can you take a look at this PR (which I've already merged) and see if it fits what what you were thinking it would look like to incorporate rubocop and bixby into our projects?

Issue #67, opened 2017-04-08
Run rubocop first for CI


mark-dce:
**STORY**
As a developer, I would like my CI builds to run rubocop before rspec so that builds fail fast on style issues.

**ACCEPTANCE**
- [x] Travis runs rubocop before rspec tests

acozine:
Closed by #106.

PR #106, opened 2017-04-24
Run rubocop before rspec

bess:
Addresses #67


acozine:
Travis does the right thing, looks good.

PR #114, opened 2017-04-25
Ensure 'Share Your Work' button goes to create ETD page.

### LINKED ISSUE #69: "Share Your Work" button doesn't work on emory.curationexperts.com
**ISSUE**
The "Share Your Work" button on the home page does not seem to do anything on the demo server.

**STEPS TO REPRODUCE**
1. Visit http://emory.curationexperts.com
1. Click on the "Share Your Work" button

**EXPECTED RESULT**
* Authenticated users should be taken to their Dashboard
* Unauthenticated users should be taken to a login page

**ACTUAL RESULT**
Nothing happens when I click the button

**NOTE** I tested the same steps in my local development environment and the button works as expected.

rotated8:
Adds the index view from Hyrax, and tweaks it.

Fixes #69

### DIFF
--- a/app/views/hyrax/homepage/index.html.erb
+++ b/app/views/hyrax/homepage/index.html.erb
@@ -1,0 +1,24 @@
+<% provide :page_title, application_name %>
+
+<% if @presenter.display_share_button? %>
+  <div class="home_share_work row">
+    <div class="col-sm-12 text-center">
+      <% if signed_in? %>
+        <%= link_to main_app.new_hyrax_etd_path,
+          class: "btn btn-primary btn-lg" do %>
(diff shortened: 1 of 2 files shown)


bess:
@rotated8 Looks like your test replaces the existing one? Don't we want to test that a submission works both at /concern/etds/new AND at root_url?

rotated8:
It doesn't exactly replace the old test, but it starts from a different URL.
The `root_url` is the homepage, and since the testing user is logged in, clicking the 'Share Your Work' button should take them to `/concern/etds/new` (plus a locale), which is the `new_hyrax_etd_path` (plus a locale).

PR #124, opened 2017-04-27
Update the Solr version.

rotated8:
Force the test solr_wrapper to use 6.5.1,  released today.


bess:
👍

acozine:
Moving forward, we should be sure that upgrading solr-wrapper only happens as part of a larger effort that also upgrades the production environment (once tests pass and everything is happy). Production Solr may lag the latest release unless there are features or performance improvements we need or want.

Issue #79, opened 2017-04-10
Decide how to handle admin set setup


mark-dce:
**ISSUE**
At present you need to run a rake task for each new environment.  We either need to add this to all set up instructions, or automate the setup of admin sets so that it can be handled via the UI

- [x] Make a system to automate the setup of admin sets
- [x] Ensure it is run whenever the application is deployed
- [x] What admin sets and workflows and user identities do we want automatically set up? 
  - [x] 4 admin sets, one for each school, just one workflow for them all, everything owned by uberadmin@localhost.com

bess:
An admin set is tied to a workflow, and the creation and use of workflows is a little complicated. I don't think we want to do this via the UI. I think it has to be automated and needs to stick with the application, more like a database migration. Also, it's a prerequisite for having an approval workflow, so I'm moving it up on the priority list.

mark-dce:
Agreed, the UI part of the description might have been a red-herring from trying to transcribe the discussion as it was happening.

There are other types of seed information that we've needed in various past projects.  One of the patterns we've used is to check the seed configuration into the repo, have some form of initializer check for the existence of the the type of configuration expected (eg. is there more than one admin set - we probably expect one for each school) and if the check fails to load the seed data into the environment in question.

As an example use case:  I've configured an admin set per school, I run rake solr_clean + rake fc_repo clean to clear out test data I've been messing around with.  I still want my admin sets configured in a consistent manner without having to run multiple rake tasks.  I also want everyone else's dev, test, and production environments to have the same predictable configuration...

PR #131, opened 2017-05-04
Automate admin sets

bess:
Addresses #79 
This uses an initializer to create four AdminSets, and give them each the one step mediated deposit workflow. These admin sets are owned by a fake user, "uberadmin@localhost.com" whose password is hardcoded. Obviously we'll need to change that when we have shibboleth auth working, but I think the mechanism is good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,7 @@ rvm:
   - 2.3.3
   - 2.3.4
 script:
+  - RAILS_ENV=test bundle exec rake db:setup --trace
+  - RAILS_ENV=test bundle exec rake db:migrate --trace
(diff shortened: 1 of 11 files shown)


bess:
@mark-dce Does this fulfill what you were looking for in terms of AdminSet setup? 
- it runs via an initializer, so you don't have to remember to run it
- it will run whenever the app is started
- it will ensure these four AdminSets exist and have the mediated deposit workflow, but it won't overwrite any other AdminSets that have been created

Note that coverage went down because I added some lines to `ability.rb` which are required for the use of hydra-role-management. I don't consider this code that needs a test, but it did drop the overall code coverage number.

PR #185, opened 2017-05-16
Config schools

alexBLR:


config/emory/schools.yml
```diff
@@ -0,0 +1,351 @@
+schools:
+  Laney Graduate School:
+    Divisions:
+      Humanities:
+        programs:
+          - name : Art History
+            id: ARTHISTMA
+          - name : Comparative Literature - MA
```
bess: Shouldn't this just be "Comparative Literature"?

bess:
@alexBLR This looks great, thank you! 

I made one comment (I think it's a typo). Also, would you please squash these three commits into one to keep our commit history easier to read? As soon as you do that I'll merge this.
Thanks!

config/emory/schools.yml
alexBLR: Yep. Let me correct that really quick

PR #218, opened 2017-05-22
Metadata agencies

alexBLR:


app/forms/hyrax/etd_form.rb
```diff
@@ -8,6 +8,16 @@ class EtdForm < Hyrax::Forms::WorkForm
     self.terms += [:department]
     self.terms += [:school]
     self.terms += [:degree]
+    self.terms += [:partnering_agency]
+
+    # if singular comes up in the future
```
bess: Hi, @alexBLR . Let's remove this method, actually. I think it might cause confusion if someone encounters it down the road. If a field needs to become singular, it should be added to the self.single_valued_fields array just below this method. That will better match the Hyrax community pattern.

bess:
Hey, @alexBLR. If you remove the commented out method on etd_form.rb, resolve your conflicts, and merge your commits, this will be good to go. Thanks!

PR #231, opened 2017-05-24
Improved workflow setup

### LINKED ISSUE #236: Superusers can manage any AdminSet
- [x] Superusers should be loaded from a config file
- [x] Superusers can edit, view, and approve in any AdminSet

bess:
* Load superusers from a config file
 * Call them "superusers" instead of "uberadmin"
 * Give superusers all roles in all workflows
 * Allow any registered user to deposit into any AdminSet

Obviously in real life we won't hard code passwords or check our production config file into a public repo. This is just to give an idea of what the config file might look like and start a conversation about how that might be managed.

Fixes #236 and #237

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ or cherokee rose is the <a href="https://georgia.gov/georgia-facts-and-symbols">
 1. Run the test suite  
     `bin/rails ci`
 
+## User and workflow setup
+
+Each Emory school has its own AdminSet, which determines the approval process for that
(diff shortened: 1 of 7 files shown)


bess:
Aargh! This is so close to working! I've got one failing test that passes in isolation, and randomly passes in the build, so it's got to be something that one of the other tests is doing.

bess:
@jenlindner Take another look at this when you get a chance? It's much improved from yesterday. Thanks!

jenlindner:
It's in great shape!! Getting all the pieces into place for this was inspiring to witness. I'm merging it.

Issue #196, opened 2017-05-17
Produce and test workflow for Laney


bess:
- [x] The workflow should be auto-populated with approvers who are loaded from a config file
- [x] Two step approval: review and approval 
- [ ] Depositor can see work in the pending_review state (How important is this? See ticket #275)
- [x] Depositor can edit work that has been sent back for edits
- [x] Depositor cannot edit or delete a work that has been approved
- [x] "Work has been reviewed" notifications go to depositor and approvers
- [x] "Work has been approved" notifications go to depositor
- [ ] "Work has been approved" notifications go to appropriate Emory faculty
- [x] Superuser can edit work in any state (in progress)
- [ ] Superusers can purge a work from any state (currently blocked by a bug. See issue #274)
- [ ] We have a workflow step for "sent to ProQuest" (see ticket #290)
- [ ] Superusers can mark a work "Dark" at any state (see ticket #291)
- [x] The Laney admin set should be set to use this workflow

The workflow should implement this diagram:  https://docs.google.com/drawings/d/1oCNgx4XiDzZdsomnTkP4LVkVBtr0rHJ4OJhZUXlmJo0/edit

fpici:
Hi Bess. I placed a Ticket to our Systems folks 5.17 afternoon for your requested "current config file for the School Reviewer’s Permissions." I may have something for you by the end of the day. :-) fran

fpici:
Added here to box 5.23.2017: https://emory.box.com/s/fnhitfsng9m4il4utc7gv50sywdqb006

PR #249, opened 2017-05-26
Load approvers from config files

bess:
Applies to #196


bess:
The build that is failing is failing because our master branch is pinned to hyrax 1.0.0, which has been pulled because it was incorrectly released, so the CI server can't install 1.0.1.

bess:
Reopening a fresh one

PR #254, opened 2017-05-26
Load approvers for each school from a school-specific config file

bess:


bess:
Closing in favor of more recent PR "Load approvers from a config file and make workflow setup idempotent"

PR #256, opened 2017-05-26
Bug fixes partnering agencies

alexBLR:


bess:
@alexBLR Glad to see this is passing now! I'll be glad to merge it with a cleaned up commit history. Maybe a new PR against master?

PR #278, opened 2017-06-02
notifications and feature test for laney workflow

bess:
Addresses #196


bess:
I believe this PR drops code coverage percentage because it contains several instances of things that are just messages. I actually am testing them, in the attached feature spec, but it doesn't show up in the coverage metrics.

PR #286, opened 2017-06-06
Overwrite hyrax save js

jenlindner:


jenlindner:
This PR requires developers to install a dependency for capybara-webkit locally: https://github.com/thoughtbot/capybara-webkit/wiki/Installing-Qt-and-compiling-capybara-webkit

PR #288, opened 2017-06-07
Superusers need to be in the approving role

bess:
Superusers need to be in the approving role in order to participate in the approval workflow


jenlindner:
Looks great!

PR #311, opened 2017-06-12
adding tests to committee fields

alexBLR:
It seems that other things are already added. I just added tests for the models.


bess:
@alexBLR I'm closing this because it's superseded by the work I just did making faculty its own object.

PR #315, opened 2017-06-13
Add title of approved work to approval notification

bess:


bess:
@jenlindner This fixes the bug you found. Thanks!

Issue #318, opened 2017-06-13
Ensure we send the right degree codes to ProQuest


shanitajohn:
**TASK** 
Review the degree.xls provided by ProQuest and: 

- [ ] Highlight degrees offered by Emory (@bnash and @fpici)
- [ ] Add any Emory degrees not in the xls (@bnash and @fpici)
- [ ] Ensure the [degree.yml](https://github.com/curationexperts/laevigata/blob/5de0035cb1e4ad0c70030606cd79e0ebfed7539d/config/authorities/degree.yml) file in laevigata is recording options that will be valid for ProQuest (@alexBLR)

**FILE**
https://emory.box.com/s/e6zpwwci4d9w0b1ekk75zt5p8d1zq33c

shanitajohn:
Hi @fran, are you the right person to complete this task? Otherwise, please redirect me. Thanks!

fpici:
@bnash108 Hi Shanita. I am passing this over to Bethany. She may be able to help with assigning this task. Hope this helps,

bnash108:
@shanitajohn I have a question in to Jay here to see if it would be possible to export all the degree codes from current ETDs. If that is possible, would a list like that meet the needs of this ticket?

shanitajohn:
@bnash108 I think that would satisfy the request but I'd like to confirm that with @bess first.

shanitajohn:
@bnash108 I asked Bess, she's not sure that list would be enough because we don't know if there's anything happening behind the scenes to resolve to what ProQuest is expecting. But, if you are able to get it from Jay (and if it's simpler to do than highlighting) please go ahead! It's always helpful to see the what is currently being used.

bnash108:
@shanitajohn I emailed my ProQuest contact for more details about their degree code list. Will report back once I hear from him.

bess:
@bnash108 I believe we already have the list of degree codes that ProQuest will accept. We just don't want to present the students with that full list -- it's very long and most of the degree options are irrelevant to them. Instead, we'd prefer to present the students with only the degrees that are actually options for them, and then we'll cross-reference those with the ProQuest list and ensure we're recording them with the ProQuest terminology (e.g., "B.A." instead of "BA").

bnash108:
[...]

PR #323, opened 2017-06-14
School dropdowns

alexBLR:


app/assets/javascripts/app.js
```diff
@@ -7,3 +7,35 @@ Hyrax.editor = function() {
         new EtdEditor(element)
     }
 }
+
+jQuery(document).ready(function () {
```
jenlindner: Hey Zo, this would be better in its own js file, for clearer separation of features. And it's a Samvera convention to put functions that need to be called when the page loads into the Blacklight onLoad function, like this: `Blacklight.onLoad(function() {}); ` .

app/views/hyrax/base/_form_my_pdf.html.erb
```diff
@@ -16,14 +16,14 @@
            <span class="btn btn-success fileinput-button">
                <span class="glyphicon glyphicon-plus"></span>
                <span>Add files...</span>
-               <input type="file" name="files[]" multiple>
+               <input id="etd_files" type="file" name="files[]" multiple style="z-index: 999999999;">
```
jenlindner: Doesn't seem like this should be necessary? And we want to avoid using inline styles.

app/views/records/edit_fields/_submitting_type.html.erb
```diff
@@ -3,4 +3,4 @@
     collection: submitting_service.select_all_options,
     include_blank: true,
     input_html: { class: 'form-control' },
-    label: "I am submitting" %>
+    label: "I Am Submitting My" %>
```
bess: Please change to "I am submitting my". The addition of "my" is good, but we don't want to title case this.

config/authorities/degree.yml
```diff
@@ -0,0 +1,25 @@
...
+    - id: 'MPH'
+      term: 'MPH'
+    - id: 'MSPH'
+      term: 'MSPH'
+    - id: 'BA'
+      term: 'BA'
+    - id: 'BS'
+      term: 'BS'
+    - id: 'BBA'
+      term: 'BBA'
```
bess: Let's move forward with this as-is for now, but note we will eventually need to align this file with the degree codes required by ProQuest. I added that task to ticket #318

config/authorities/emory_college_programs.yml
```diff
@@ -0,0 +1,129 @@
...
+    - id: POLISCI
+      term: Political Science
+    - id: PSYCHBA
+      term: Psychology
+    - id: PSYCLINGBA
+      term: Psychology and Linguistics
+    - id: SOCBA
+      term: Sociology
+    - id: WOMENSTBA
+      term: Women's Studies
```
bess: Why do some of these files have the same values for id and term, and others have a code for id? I understand that's probably because that's what was in the file we were given, but do we ever want to record the code instead of the term? Seems to me we will always want a human readable and searchable string to be the recorded value. Thoughts?

spec/features/rollins_workflow_etd_spec.rb
```diff
@@ -14,19 +14,21 @@
       ActiveFedora::Cleaner.clean!
       w.setup
       login_as user
+      visit("/concern/etds/new")
+    end
+    # weak test of the fields but it was failing due to webkit issues with attach_file
+    scenario "Miranda submits school and department", js: true do
+      select("Rollins School of Public Health", from: "School")
+      select("Global Health", from: "Department", match: :first)
```
bess: I think testing the drop down menus separately is a fine solution for now.

spec/models/etd_spec.rb
```diff
@@ -132,6 +148,42 @@
...
+    context "with an existing ETD that has a committee members field defined" do
+      subject do
+        described_class.create.tap do |etd|
+          etd.committee_members = committee_members
+        end
+      end
+      its(:committee_members) { is_expected.to eq(committee_members) }
+    end
+  end
+
```
bess: Would you please remove these tests for committee_chair and committee_members? We need those fields to be an instance of a Faculty object (which has a name, affiliation, and netid), not a simple string. While it is allowing you to set committee_chair and committee_members to a string, it shouldn't be.

config/authorities/emory_college_programs.yml
alexBLR: Some of them don't have ids, so I made them up for now. Then, others have two ids so I had no idea what to do with these.

bess: I would remove all the ids, and if you need to have an id and a term then just make them identical (e.g., id: sociology, term: sociology), the way it is for some of the files already. I think we want to be recording the actual value, not the code.

Issue #292, opened 2017-06-07
Update research fields with latest data from ProQuest


bess:
- [ ] Research fields file updated with latest data from ProQuest (@bnash has asked for the latest data)

bnash108:
[ProQuestSubjectCategories_06-12-2017.xlsx](https://github.com/curationexperts/laevigata/files/1068787/ProQuestSubjectCategories_06-12-2017.xlsx)

bnash108:
@bess @shanitajohn ProQuest provided updated list of research fields above

shanitajohn:
@rotated8 Please use the XLS that Bethany provided above to complete your work on the research fields drop down. Thanks!

rotated8:
@bnash108 The spreadsheet appears to have problems, which you can see better in github: https://github.com/curationexperts/laevigata/pull/335/files. Can we ask ProQuest for fixes?

bnash108:
[ProQuest_subjects_06-19-2017.xlsx](https://github.com/curationexperts/laevigata/files/1086234/ProQuest_subjects_06-19-2017.xlsx)

bnash108:
@rotated8 Above is the updated spreadsheet from ProQuest
Let me know if I need to circle back to ProQuest with questions

Issue #188, opened 2017-05-16
Create research field drop-down menu


shanitajohn:
**STORY**
As a depositor, I would like to be able to select my research field from a list of available ProQuest options so that in the future, when researchers browse work by topic or fields, they will discover my work as well. 

**ACCEPTANCE CRITERIA**
- [x] I would like to be able to click on the drop-down and see a list of ProQuest research fields 
- [x] I would like to be able to choose a research field by clicking on the name in the list
- [ ] I would like to be able to designate a research field as "primary"
- [x] I would like to be able to add Additional Research fields (up to 2)
- [x]  I would like to be able to remove an Additional Research field 
- [ ] Since I can add up to 3 research fields, I would like a way to see which field I have designated as my primary research field 

**WIREFRAMES (updated 6/2/17)**
https://emory.box.com/s/l2gsln7vzsl972r0x8pe5i305f3h8ktf

Sample config file: 
https://emory.box.com/s/877m5ouglzg8a7sq383a5x2one5bgi4b

shanitajohn:
Related to Ticket #242

rotated8:
Parts of this were implemented in #277

rotated8:
Config file in app: https://github.com/curationexperts/laevigata/blob/about-me-UI/config/authorities/research_fields.yml

shanitajohn:
@rotated8 Can you add a "size" to this ticket please? Thanks!

rotated8:
I'm not sure how to size this ticket, because I'm not sure how to approach the primary designation. Thoughts, @bess ? Limiting to three fields is a 1 for the backend, not sure about the frontend.

bess:
I took the primary designation out of this ticket, since the research drop down menu is otherwise done. Let's regroup over in ticket #365

PR #334, opened 2017-06-15
Store research field names and ids

### LINKED ISSUE #309: Split Research Field metadata into two fields
- [ ] research_field_id should store the proquest id of the research field selected (predicate TBD, ask Emily)
- [ ] research_field should store the human readable string
- [ ] research_field should use http://purl.org/dc/elements/1.1/subject instead of dcterms:subject

rotated8:
Fixes #309 

The terms and ids have switched in the service and tests. This is necessary to look up a field's code by the name of the field. Because of this, the helper and renderer can be removed, and we don't need to look up labels for tests. Now, we need the service's #select_all_ids method, so the dropdown shows and stores the names of the research fields.

The model validates three or fewer fields, as per #188, but does not address marking one of them as primary.

The research_field_id attribute is set when the ETD is saved, looking up the fields' codes at that time.

The predicate for research_field_id is the temporary one given by Emily, who will be out until Tuesday. The predicate for research_field is DC11 instead of dcterms.

Changing the fields to the ones given by ProQuest (#292) will happen separately, as I think ProQuest gave us a bad file.


rotated8:
CI failures are the same as the failures on master.

PR #342, opened 2017-06-16
Add embargo to legacy metadata fixture

bess:


bess:
It doesn't make sense that this change is dropping coverage that much. I'd like to merge it anyway, please.

Issue #306, opened 2017-06-09
Basic ETD show page


bess:
- [x] Wireframe for display screen
- [x] Display all the fields for an ETD

shanitajohn:
**WIREFRAME:** https://emory.box.com/s/wdiz63c6tntt08hk6v158hb5wnh65m8r

bess:
This is done, but for metadata fields only. Still required is Files and FileSet metadata. Tracking that work under #359 and #360

Issue #166, opened 2017-05-11
Display complete metadata for a legacy ETD


bess:
Make sure we have all the metadata fields completely mapped by being able to display one legacy Emory ETDs in Laevigata. Use ateer_Laney from https://app.box.com/folder/23319576726

** Acceptance Criteria**
- [x] Display all required metadata on the show page for the ETD

bess:
This is done, for the metadata only. Still needed is the display of associated files and file set metadata, and the display and enforcement of embargoes. I've opened tickets for that work under #359 and #360

PR #356, opened 2017-06-19
Committee, abstract, toc

bess:
Add committee chair and committee member to solr index and show view
Add abstract and table_of_contents to model, index, fixture factory, and show view

Addresses #166 and #306


jenlindner:
NICE!!!!!

PR #358, opened 2017-06-20
Primary supplemental metadata

alexBLR:


app/models/etd.rb
```diff
@@ -79,6 +79,50 @@ def hidden?
...
+  property :keyword, predicate: "http://schema.org/keywords" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :file_format, predicate: "http://purl.org/dc/elements/1.1/format" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :identifier, predicate: "http://purl.org/dc/terms/identifier" do |index|
+    index.as :stored_searchable, :facetable
```
bess: We are already inheriting identifier, with this predicate, from Hyrax. Let's not override it.

app/models/etd.rb
```diff
@@ -79,6 +79,50 @@ def hidden?
...
+  end
+
+  property :author, predicate: "http://id.loc.gov/vocabulary/relators/aut" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :description, predicate: "http://purl.org/dc/elements/1.1/description" do |index|
+    index.as :stored_searchable, :facetable
```
bess: description should not be facetable

app/models/etd.rb
```diff
@@ -79,6 +79,50 @@ def hidden?
...
+  property :file_format, predicate: "http://purl.org/dc/elements/1.1/format" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :identifier, predicate: "http://purl.org/dc/terms/identifier" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :author, predicate: "http://id.loc.gov/vocabulary/relators/aut" do |index|
+    index.as :stored_searchable, :facetable
```
bess: We already have the "creator" field stored with this predicate. Since "creator" is the one that Hyrax expects and is built around, let's use that unless we have a strong reason to do something different.

app/models/etd.rb
```diff
@@ -79,6 +79,50 @@ def hidden?
...
+  end
+
+  property :description, predicate: "http://purl.org/dc/elements/1.1/description" do |index|
+    index.as :stored_searchable, :facetable
+  end
+
+  property :file_type, predicate: "http://purl.org/dc/elements/1.1/format" do |index|
```
bess: Is this the file_type for the primary file? Or secondary? It isn't clear from this file name. It has to be secondary, right? Since primary will always be PDF?

bess: Also, should this field exist on the primary ETD file? Shouldn't it be stored with the supplementary file in question?

bess:
Let's close this and if we want to pull small pieces out elsewhere we can do that.

PR #366, opened 2017-06-20
Require factory_girl in production so we can load fixture objects for…

bess:
… demos in production


bess:
Okayed by @acozine in slack

PR #374, opened 2017-06-21
Emory theming

alexBLR:


alexBLR:
The build fails because Travis is unable to connect to Redis. Very weird.

PR #408, opened 2017-06-29
Enforce embargoes for file download

### LINKED ISSUE #359: Embargoes on ETD show page
- [x] Honor legacy embargo data
- [x] Clearly show what is under embargo and when it will be available
- [x] Ensure superusers can see embargoed content
  - [x] For abstract
  - [x] For table of contents
  - [x] for file download

bess:
Make an ETD Factory that can attach files as expected
Add a rake task that can build sample data

Closes #359

### DIFF
--- a/app/views/hyrax/base/_items.html.erb
+++ b/app/views/hyrax/base/_items.html.erb
@@ -1,0 +1,19 @@
+<h2><%= t('.header') %></h2>
+<% if presenter.member_presenters.present? %>
+  <table class="table table-striped related-files">
+    <thead>
+      <tr>
+        <th><%= t('.thumbnail') %></th>
+        <th><%= t('.title') %></th>
+        <th><%= t('.date_uploaded') %></th>
+        <th><%= t('.visibility') %></th>
+        <th><%= t('.actions') %></th>
+      </tr>
+    </thead>
+    <tbody>
+      <%= render partial: 'member', collection: presenter.member_presenters %>
+    </tbody>
+  </table>
+<% elsif can? :edit, presenter.id %>
+    <div class="alert alert-warning" role="alert"><%= t('.empty', type: presenter.human_readable_type) %></div>
+<% end %>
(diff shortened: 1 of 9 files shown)


bess:
Looks like travis is having issues. AFAIK, this build passes. Let's give it a bit and see if travis recovers.

spec/factories/etd.rb
```diff
@@ -11,6 +12,14 @@
       abstract { [] << FFaker::Lorem.paragraph }
     end
 
+    factory :etd_with_pdf do
```
rotated8: I don't see this used in the new code. Is this for upcoming tests? How does the pdf get attached?

spec/factories/uploaded_file.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :uploaded_file, class: Hyrax::UploadedFile do
```
rotated8: Is this factory being used? Could the EtdFactorySpec use it?

app/views/hyrax/base/_items.html.erb
```diff
@@ -0,0 +1,19 @@
+<h2><%= t('.header') %></h2>
```
rotated8: I see this is copied from Hyrax. What is changing?

bess: Nothing is changing yet, but it will need to. I probably copied this over prematurely, but figured I'd need it for the next thing I'm doing so I left it in place.

spec/factories/uploaded_file.rb
bess: It isn't be used yet, but I made it as I was figuring out how to attach files and I figured I will end up using it for future work, so I left it in place.

spec/factories/etd.rb
bess: I had a notion I was going to make a factory to make an ETD with a PDF attached, but the attach file to work job wouldn't run in this context. I left it in place thinking I'd use it later. Happy to remove it to de-clutter.

rotated8:
Sounds like everything will get used in the future. I'm 👍 once Travis 💚s it.

bess:
Closing in favor of https://github.com/curationexperts/laevigata/pull/410

Issue #427, opened 2017-07-05
Not all chairs and committee members are faculty


bess:
Instead of "Emory / Non-Emory Faculty" say 
  - [x] "Emory / Non-Emory Committee chair"
  - [x] "Emory / Non-Emory Committee member"

PR #439, opened 2017-07-06
Changed text because not all committee members are faculty. Story #427

val99erie:


val99erie:
Don't merge yet.  I'll need to rebase and fix conflicts.  I'll add another comment when it's ready to go.

app/assets/javascripts/about_me_and_my_program.es6
```diff
@@ -10,7 +10,7 @@ export default class AboutMeAndMyProgram {
 
   attach_committee_select_listeners(selector){
     $('.about-me').on('change', selector, function(){
-      if ($(this).val() == "Emory Faculty"){
+      if ($(this).val() == "Emory Committee Chair" || $(this).val() == "Emory Committee Member"){
```
val99erie: @jenlindner I don't necessarily want to do it as part of this PR, but I'm wondering if we can pass in the possible values from the ruby to the javascript, so that we only need to define those values in one place.

jenlindner: Yeah, it would be nice, not a high priority before the launch, but post.

PR #450, opened 2017-07-10
mark shib accounts

bess:
We have to mark shibboleth accounts as belonging to shibboleth.
@acozine This should fix the problem Bethany is having immediately. It will take a little longer to switch us over to using uid for account management.


lib/workflow_setup.rb
```diff
@@ -101,13 +103,25 @@ def admin_role
...
+    admin_role.save
+    @superusers_config["superusers"].keys.each do |provider|
+      @superusers_config["superusers"][provider].each do |s|
+        make_superuser(s, provider)
+      end
+    end
+  end
+
   # Make a superuser
   # @param [String] the ppid of the superuser
```
rotated8: Should we add the provider param here?

bess:
Superseded by #453

PR #460, opened 2017-07-11
Update user model to match new shibboleth AuthHash structure

bess:
Fixes an issue where Emory folks can't authenticate with shibboleth.


acozine:
Yes, this does the correct translation - in the devise initializer we have `uid_field: serialNumber` so in the user model we need `user.ppid = auth.uid` to store the serialNumber from Shibboleth as the local field `ppid`.

PR #470, opened 2017-07-12
Feature spec for creating an ETD

val99erie:
It still needs more work, but this is a first pass at getting an end-to-end feature spec working.

Added rubocop exclusions because this code is much more readable than what rubocop wanted.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -26,6 +26,12 @@ RSpec/MultipleExpectations:
 Style/FileName:
   Enabled: false
 
+Style/Semicolon:
+  Enabled: false
+
+Layout/MultilineMethodCallIndentation:
+  Enabled: false
+
 RSpec/InstanceVariable:
   Enabled: false
 
--- a/spec/features/create_etd_spec.rb
+++ b/spec/features/create_etd_spec.rb
@@ -1,5 +1,4 @@
-# Generated via
-#  `rails generate hyrax:work Etd`
+# Generated via `rails generate hyrax:work Etd`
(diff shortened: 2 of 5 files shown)


val99erie:
The Travis build is failing intermittently.  It looks like the ETD was being suppressed by the approver workflow, even though the show page should be visible to the user who created the ETD.  (In the error message, the show page has the text "The work is not currently available because it has not yet completed the approval process").  I'm wondering if it's a thread problem.

I'm about to log out for the evening.  I'll try to dig into it tomorrow.

bess:
It's a known issue that the user who just submitted a work can't see it again until it's approved. Let's check in on whether we should prioritize fixing that, or whether it would be better to check the show screen as the approver.

val99erie:
@bess I don't think that's it.  The failure is only intermittent, and I can't get it to fail on my dev env.

val99erie:
The build fails intermittently, and when it fails, it has a different workflow state than when it passes.

When the build fails:
```ruby
workflow state: pending_approval
State: http://fedora.info/definitions/1/0/access/ObjState#inactive
suppressed?: true
```

When the build passes:
```ruby
workflow state: deposited
State: http://fedora.info/definitions/1/0/access/ObjState#active
suppressed?: false
```

In both cases, the `etd.depositor` matches the logged in user's `ppid`.

val99erie:
Don't merge yet.  Waiting for code from @bess in another PR.

val99erie:
@bess Ok, I think this is finally ready to go.  Please review and merge.

PR #497, opened 2017-07-18
Set initial pre-graduation embargo

### LINKED ISSUE #462: Set placeholder embargo date
**ACCEPTANCE CRITERIA**

- [ ] Upon submission embargo date is set to submission_date + 6 years. 
- [ ] We know the embargo length 
- [ ] Show page displays embargo length and what has been embargoed. 
- [ ] A future automated job, when checking graduation status, updates the embargo date to be graduation_date + embargo length

bess:
Fixes #462

### DIFF
--- a/app/actors/hyrax/actors/interpret_visibility_actor.rb
+++ b/app/actors/hyrax/actors/interpret_visibility_actor.rb
@@ -1,0 +1,67 @@
+# See `README_embargos.md` for an explanation of how Laevigata embargoes
+# differ from the usual Hyrax behavior
+module Hyrax
+  module Actors
+    class InterpretVisibilityActor < AbstractActor
+      def create(attributes)
+        @attributes = attributes
+        save_embargo_length
+        attributes.delete(:embargo_release_date)
+        attributes[:visibility] = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PUBLIC
+        apply_pre_graduation_embargo
+        curation_concern.save
+        next_actor.create(attributes)
+      end
+
+      def update(attributes)
+        @attributes = attributes
+        save_embargo_length
+        attributes.delete(:embargo_release_date)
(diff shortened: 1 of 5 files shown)


app/actors/hyrax/actors/interpret_visibility_actor.rb
```diff
@@ -0,0 +1,46 @@
+module Hyrax
+  module Actors
+    class InterpretVisibilityActor < AbstractActor
```
val99erie: This class is modified from Hyrax, right?  I feel like we need some documentation about why this class was copied over, and how it differs from the same class in Hyrax.

In general (not necessarily as part of this PR, but maybe as a separate story), I think we need a master document that describes how the embargoes are supposed to work in laevigata.  Because it's different from standard hyrax embaroes, this is an area that could cause a lot of confusion.  Maybe we could add a `README_embargo.md` or add a `docs` directory to the project.

spec/actors/hyrax/actors/interpret_visibility_actor_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+          "department" => ["Divinity"],
+          "files_embargoed" => "true",
+          "abstract_embargoed" => "false",
+          "toc_embargoed" => "false"
+        }
+      end
+
+      context 'with an embargo length' do
+        let(:date) { Time.zone.today + 2 }
+        it 'sets a temporary embargo release of date six years from today' do
```
val99erie: Why is there a temporary embargo date?  Why is it 6 years?  Does the original chosen value of 6 months get saved anywhere or used for anything?

val99erie: ah, I see where the 6 months gets stored

bess:
@val99erie Take another look? I added a README as suggested.

Issue #474, opened 2017-07-13
When I save the form I should see all of the content on the show page


shanitajohn:
**ACCEPTANCE CRITERIA** 
- [x] All the fields I fill in should appear on the show page according to the layout of the wireframes 
- [x] For fields that are not displayed on the show page (for example post-grad email) check the Rails console for the Fedora console to make sure the data's there

shanitajohn:
Related to #462

val99erie:
Known broken fields:
- [ ] committee members
- [ ] committee chair
- [ ] post-grad email

val99erie:
Still left to investigate:
- [ ] Abstract
- [ ] Table of contents

PR #504, opened 2017-07-20
required committee members

### LINKED ISSUE #393: make Committee Members required when schools that require them are selected
**ACCEPTANCE CRITERIA**
- [x] If I select a school that doesn't require committee members and I leave the committee members blank, I should be able to successfully save the form with no errors
- [x]  If I select a school that does require committee members and I leave the committee member fields blank, and I save, I do not see a green checkmark and I the save button on the review page is grayed out

alexBLR:
Fixes #393


bess:
Closes #393

app/assets/javascripts/dependent_dropdowns.js
```diff
@@ -17,8 +17,16 @@ observer.attr('disabled', true);
 observed.on('change', function () {
 observer.empty().append(prompt);
 if (observed.val()) {
-// console.log(observer_dom_id)
-// console.log(observed.val())
+
+if($('#etd_school option:selected').index()==4){
```
jenlindner: I think this would be better with a check against the value of the option, to make it clear that that particular school has the special condition.

alexBLR: thank you, Jen.

spec/features/create_etd_spec.rb
```diff
@@ -58,8 +58,8 @@
       select 'CDC', from: 'Partnering agency'
       select 'PhD', from: 'Degree'
       select 'Dissertation', from: 'I am submitting my'
+      fill_in 'Committee Member', with: 'John Smith'
```
val99erie: You don't need to fill this field in because this doesn't actually test anything.  The committee members field isn't working on the back end yet, so this part isn't quite ready yet.  I'll add in something like this as part of story #474.

alexBLR: Thank you! 👍

jenlindner:
Hey @alexBLR, I realize I didn't give you the best information here, I apologize. I think that part of @little9's work entails adding 'optional' flags to the tabs that are optional (and 'required' flags to the required ones, I believe). And that keeping the UI consistent with these fields would mean displaying an 'optional' flag next to them in every case but when Rollins is selected. I think that would be better, what do you think, @little9?

PR #505, opened 2017-07-20
Refactor workflow feature specs to only test workflow, not ETD submis…

bess:
…sion

### DIFF
--- a/spec/features/candler_workflow_etd_spec.rb
+++ b/spec/features/candler_workflow_etd_spec.rb
@@ -5,48 +5,26 @@ require 'active_fedora/cleaner'
 require 'workflow_setup'
 include Warden::Test::Helpers
 
-RSpec.feature 'Create a Candler ETD' do
-  let(:user) { create :user }
+RSpec.feature 'Candler approval workflow' do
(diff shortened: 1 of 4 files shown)


bess:
@alexBLR If you re-run your PR after this one is merged, I believe it should pass.

spec/features/candler_workflow_etd_spec.rb
```diff
@@ -5,48 +5,26 @@
...
+  let(:etd) { FactoryGirl.create(:sample_data, school: ["Candler School of Theology"]) }
   context 'a logged in user' do
     before do
+      allow(CharacterizeJob).to receive(:perform_later) # There is no fits installed on travis-ci
       ActiveFedora::Cleaner.clean!
       w.setup
-      login_as user
-      visit("/concern/etds/new")
+      actor = Hyrax::CurationConcern.actor(etd, ::Ability.new(depositing_user))
+      actor.create({})
```
jenlindner: This looks great. Could you maybe add a comment about why the actor gets created here? I'm new to this part of the stack and don't understand why it's needed because it seems like creating the etd would be all you'd need.

PR #512, opened 2017-07-21
review my etd preview and submit

jenlindner:
@bess, would you mind pulling this down and checking it out? And @val99erie I marked your create etd spec as pending rather than adapt it to the new review but if you rebase your changes will overwrite it and you can adapt it where you'd like to.


jenlindner:
I found that the embargo validation wasn't getting triggered by change events on its selects, and added that, force-pushed to kick off the build again.

bess:
Yes! This looks great!!

jenlindner:
Thank you!!

PR #528, opened 2017-07-24
Suppress partners, required committee, tinymce

alexBLR:


bess:
@alexBLR I'd expect to see the TinyMCE fix in a separate branch with a separate PR. It's not dependent on the suppressed partners or the required committee members work, right? Each of these should be separate branches and separate PRs, please.

PR #576, opened 2017-08-01
Metadata supplemental files 

jenlindner:
Hi @bess, this is a very very rough working version of what we talked about earlier, more or less. It creates a structure like this `"supplemental_file_metadata"=>{"0"=>{"filename"=>"5-MySuppFiles.png", "title"=>"foo"}, "1"=>{"filename"=>"9-ReviewSubmit_3.png", "title"=>"bar"}, "2"=>{"filename"=>"8-ReviewPreview_2-1.png", "title"=>"baz"}}`. I'll continue working on it but just wanted to give you a chance to look at it, see what you think, see if it raises any red flags.


bess:
@jenlindner Yes, I think that will work! :D

jenlindner:
@bess Thank you!

val99erie:
@jenlindner Could you remove that OSX file from the PR `spec/.DS_Store`

spec/features/upload_supplemental_files_spec.rb
```diff
@@ -0,0 +1,160 @@
...
+
+      expect(page).to have_content('File Name')
+      expect(page).to have_content('Title')
+      expect(page).to have_content('Description')
+
+      expect(page).to have_select('supplemental_file_file_type', options: ['Text', 'Dataset', 'Video', 'Image', 'Sound', 'Software'])
+
+      expect(page).to have_link('Hide Additional Metadata')
+
+      sleep(5)
```
val99erie: Instead of `sleep(5)`, please use something like `wait_for_ajax(5)` so that it won't sleep for the full 5 seconds if it doesn't need to.

Another way to do it would be to add an `expect` statement here that is expecting a new element to show up on the page.  I can help you with that if you don't know what I mean.  But `wait_for_ajax` will probably work just fine.

app/controllers/hyrax/etds_controller.rb
```diff
@@ -10,6 +10,7 @@ class EtdsController < ApplicationController
     self.show_presenter = EtdPresenter
 
     def create
+      apply_supplemental_file_metadata(params) if params["etd"]["supplemental_file_metadata"]
```
val99erie: Is this behavior only supposed to happen when creating an ETD?  What happens when a user edits an existing ETD?  Are they supposed to be able to edit the supplemental metadata?  (Is create vs edit in scope for this PR?)

spec/features/upload_supplemental_files_spec.rb
jenlindner: yeah, wait_for_ajax() is better, since we can set its max, the expects will just wait until the default.

app/controllers/hyrax/etds_controller.rb
jenlindner: It isn't, just create is.

PR #581, opened 2017-08-02
Fixed feature spec to create an ETD

val99erie:


val99erie:
It looks like capybara is crashing.  I'll mark this as DO NOT MERGE, and I'll come back and troubleshoot after I finish my current story.

PR #587, opened 2017-08-06
Use narrower exclusion in rubocop.yml

mark-dce:


mark-dce:
Turns out additional files need to be excluded - you can either do that manually similar to 
https://github.com/curationexperts/epigaea/commit/961c26475ca936bb5b988f618e2186d38003122f

Or, if Bixby is updated, we can just update the bixby dependency and re-enable the filename Cop.

jenlindner:
Just put in a PR to update Bixby, will re-enable fielname cop after it gets merged.

jenlindner:
Addressed by #615.

PR #597, opened 2017-08-09
Remove database_auth

### LINKED ISSUE #440: Disable database authentication in production mode
We need a way to be able to log into the app in development mode that doesn't require us to get fake Shibboleth accounts from Emory.

One possible solution would be to allow Devise's database authentication if the Rails environment is in development mode, so devs can keep using database auth.

Another possible way would be to document how to run Shibboleth on our laptops, and then each dev can maintain their own set of user logins.

In production mode, the only possible way to log in should be through Shibboleth.

- [ ] In production mode, users can only log in using Shibboleth.

- [ ] In development mode, the developer has a way to log in that doesn't require access to external Shibboleth.  (i.e. I should be able to log in on my laptop, even if I'm disconnected from the internet.)

- [ ] In test mode, same as development mode:  I should be able to run the test suite without connecting to the internet.

- [ ] For DCE's demo environment(s), we need to be able to log in somehow (Probably using DCE's Shibboleth.  See comments below)

bess:
Closes #440


bess:
Thank you, @acozine. I think we've got it, thanks to you.

PR #612, opened 2017-08-09
fixing supplemental metadata ui bug and taking most tab specs out of ci

jenlindner:


spec/features/upload_supplemental_files_spec.rb
```diff
@@ -96,6 +96,22 @@
...
 
+    scenario "checking 'no files' after uploading a few and entering metadata hides the files and metadata form", js: true unless continuous_integration? do
+      click_on('Supplemental Files')
+
+      within('#supplemental_fileupload') do
+        page.attach_file('supplemental_files[]', "#{fixture_path}/magic_warrior_cat.jpg")
+      end
+
+      click_on('Show Additional Metadata')
+      sleep(10) # ugly but works
```
val99erie: Does `wait_for_ajax(10)` work here?  If so, we should use that instead.  Or you could do something like:
`expect(page).to have_css(something)`
where "something" is an element that appears on the page after you click the 'Show Additional Metadata' button.

val99erie: The `expect` line should wait until the correct css shows up on the page.

jenlindner: It doesn't. I tried a lot of variations on expect with the max time adjusted to be 10, but none of them worked here. It's a bootstrap component that handles its own showing and hiding, that very well might have an impact, and I agree it would be better to use a wait_until, which is what the comments say. But I don't think it's worth the extra time it would take at this point.

val99erie:
Looks like this branch needs to be rebased before we can merge this PR

PR #637, opened 2017-08-15
safe the tinymce html in show view

jenlindner:
safe_output

### DIFF
--- a/app/views/hyrax/base/_abstract.erb
+++ b/app/views/hyrax/base/_abstract.erb
@@ -1,2 +1,2 @@
 <h2>Abstract</h2>
-<p><%= presenter.abstract_with_embargo_check %></p>
+<p><%= sanitize presenter.abstract_with_embargo_check %></p>
--- a/app/views/hyrax/base/_table_of_contents.erb
+++ b/app/views/hyrax/base/_table_of_contents.erb
@@ -1,2 +1,2 @@
 <h2>Table of Contents</h2>
-<p><%= presenter.toc_with_embargo_check %></p>
+<p><%= sanitize presenter.toc_with_embargo_check %></p>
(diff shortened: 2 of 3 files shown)


app/views/hyrax/base/_abstract.erb
```diff
@@ -1,2 +1,2 @@
 <h2>Abstract</h2>
-<p><%= presenter.abstract_with_embargo_check %></p>
+<p><%= presenter.abstract_with_embargo_check.html_safe %></p>
```
val99erie: Can you explain why it's necessary to mark it as `html_safe`?  Is it displaying improperly?  This is user-entered data, so if we mark is as `html_safe` we are leaving an opening for malicious data.

val99erie: Maybe try something like this:

1. In the spec, instead of `Literature from the US`, have the user input something like `Literature from the US <script>alert("attack!");</script>`

2. If you must call `html_safe` to get it to display properly, call `sanitize` first.  See this:
https://stackoverflow.com/questions/19773837/is-it-safe-to-permit-user-add-content-using-tiny-mce/26486006#26486006

jenlindner: Yes, because it was displaying all the html tags used to format the content, in the show view. Bess noticed. 

I checked how Hyrax currently handles this and it's with html_safe in views, mostly: https://github.com/samvera/hyrax/search?utf8=%E2%9C%93&q=html_safe&type=

I'll look at sanitize

jenlindner: @val99erie, did you not look at the show view when you did this ticket?

val99erie: Ah I see, Hyrax uses `safe_join` before it calls `html_safe`, so that's how they keep it secure.  Ok looks good.

Issue #648, opened 2017-08-17
Remove save buttons from application


shanitajohn:
**ACCEPTANCE**
OPTION 1
- [ ] Change the save button to take me to the next tab - i.e. change the text to "Proceed" and have the button link to the anchor for the first item on the next tab

OPTION 2
- [ ] Remove the button and put highly visible text at the bottom of the form that says something similar to "Please proceed to the next Tab to continue the submission process"

COMMON TO BOTH OPTIONS
- [ ] Add the same behavior at the bottom of all tabs (except for the last one)

val99erie:
Note: There are some feature specs that currently use those buttons.  We'll have to figure out a different way to make those work.

PR #678, opened 2017-08-21
hiding save buttons with css

jenlindner:
@bess @little9 this PR hides one of the save buttons rather than removing it only because the create review spec (the most complete test of the create feature) can't pass without it. I tried for a better solution all day with no reliable success, and while this is no one's first choice for a solution, it will work.

### DIFF
--- a/app/assets/stylesheets/etd.scss
+++ b/app/assets/stylesheets/etd.scss
@@ -99,6 +99,17 @@ ul.admin-sidebar {
   height: inherit;
 }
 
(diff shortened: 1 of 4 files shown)


bess:
@jenlindner This looks fine to me, and I tried it out locally and it works great. I'm going to merge this, but take a look at the ticket; it might have been updated since you first started working on this. (https://github.com/curationexperts/laevigata/issues/648) Mark would like some indication to the user that they should continue to the next tab.

PR #698, opened 2017-08-22
Update legacy search message and add it to top of search results screen.

bess:
Per request from @shanitajohn


fpici:
Hi Shanita. I got your message from Tiffany about text for the etd-legacy text and placement. Here is our SCO suggested text: "During our current ETD migration process, some ETD records may only be available on our legacy ETD site.  Please visit legacy-etd.library.emory.edu if you cannot find the ETD you are looking for on this new site."

I also have an idea for placement BUT I will leave that entirely up to y'all :-)  see screenshot

bess:
Updated to match @fpici 's text.

PR #758, opened 2017-09-13
Sanitize verbose ms word document formatting pasted into abstract and…

### LINKED ISSUE #756: Students receiving unknown error upon submission
Per @fpici: 
I have had two students complain about submission errors in ETDs, one today and one yesterday. The common complaint is that each student can go through the entire process with all green checks satisfied and then they when they hit Submit, they get this error page (see screenshot from student) which blows out all of their work.

bess:
… table of contents fields

Fixes #756


bess:
Requested change from @mark-dce : Expand list of allowed html elements to include all the tiny mce tags we've enabled. Use the fixture object that @jenlindner made, which lists all the valid ones.

PR #790, opened 2017-11-16
Email notifications

### LINKED ISSUE #534: System should send notifications via email
**Update description here to articulate current strategy**

Notifications for embargoes and workflow triggers should be sent via email as well as site notifications). We have a couple of options. My preference would be to route all notifications through gmail using `sendemail`. I will investigate whether this can be done with an emory no-reply address. If we can't make it work, we could run a mail server on the web server.
ACCEPTANCE:
- [ ] all notifications should be sent via email unless the user has no email address
- [ ] notifications should have a "sent from" address like `no-reply@etd.library.emory.edu`

bess:
Fixes #534


config/environments/test.rb
```diff
@@ -1,4 +1,11 @@
 Rails.application.configure do
+  config.registrar_data = ::Rails.root.join('spec', 'fixtures', 'registrar_sample.json')
+  # Where to write ProQuest export packages
+  config.proquest_export_directory = ::Rails.root.join('spec', 'fixtures', 'proquest', 'exports')
```
val99erie: If the specs are going to write files, can you put them under 'tmp' instead of under 'spec/fixtures'?

spec/mailers/previews/proquest_mailer_preview.rb
```diff
@@ -0,0 +1,8 @@
+# Preview all emails at http://localhost:3000/rails/mailers/proquest_mailer
+class ProquestMailerPreview < ActionMailer::Preview
+  let(:etd1) { FactoryGirl.create(:sample_data) }
+  let(:etd2) { FactoryGirl.create(:sample_data) }
+  def file_transfer_preview
+    ProquestMailer.file_transfer([etd1.id, etd2.id])
+  end
```
val99erie: Do you intend to add specs to this file, or should it be deleted?

bess:
I'm going to break this into pieces.

PR #798, opened 2017-11-20
Loosen restrictions for capistrano version

bess:
I'm tired of tripping on this every time there's a minor capistrano update


jenlindner:
Heh. Super cool

PR #804, opened 2017-11-27
Start moving all environment variables to dotenv

### LINKED ISSUE #802: Use dotenv for email config
- [ ] production.rb should be in version control and should reference ENV variables

bess:
Closes #802

### DIFF
--- a/dotenv.sample
+++ b/dotenv.sample
@@ -1,6 +1,6 @@
 # Copy this file to .env.development or .env.production and populate with appropriate values
 export ACTION_MAILER_HOST=localhost
-export HYRAX_HOST=**YOUR DATA HERE**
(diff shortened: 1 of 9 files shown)


dotenv.sample
```diff
@@ -0,0 +1,7 @@
+# Copy this file to .env.development or .env.production and populate with appropriate values
+export ACTION_MAILER_HOST=localhost
```
jenlindner: @bess Do we need this? We don't seem to use it.

dotenv.sample
```diff
@@ -0,0 +1,7 @@
...
+export HYRAX_HOST=**YOUR DATA HERE**
```
jenlindner: This should be RAILS_HOST, looks like.

PR #838, opened 2017-12-04
displaying copyright data to approvers

jenlindner:
Fixes #827. This code didn't make it into the release, so I'm submitting another PR. Also I think I have the spec that covers it written better, and it's passing for me locally.

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -104,4 +104,16 @@ class SolrDocument
   def rights_statement
     self[Solrizer.solr_name('rights_statement')]
   end
+
+  def copyright_question_one
+    self[Solrizer.solr_name('copyright_question_one')]
+  end
+
+  def copyright_question_two
+    self[Solrizer.solr_name('copyright_question_two')]
+  end
+
+  def copyright_question_three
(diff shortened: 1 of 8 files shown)


app/models/solr_document.rb
```diff
@@ -104,4 +104,16 @@ def research_field
   def rights_statement
     self[Solrizer.solr_name('rights_statement')]
   end
+
+  def copyright_question_one
+    self[Solrizer.solr_name('copyright_question_one')]
+  end
+
+  def copyright_question_two
+    self[Solrizer.solr_name('copyright_question_two')]
```
val99erie: I know that you are just copying the existing pattern in this file, but we should try to avoid using `Solrizer.solr_name`.  There is an email thread on samvera-tech that points out that it is a performance problem.
https://groups.google.com/forum/?hl=en#!topic/samvera-tech/uMGYsFDFyno

So, instead of:
`self[Solrizer.solr_name('copyright_question_two')]`
We should just use:
`self['copyright_question_two_tesim']`

I'll merge this PR as-is, but keep that in mind for the future.

PR #873, opened 2018-02-03
Remove remaining resque remnants

bess:
Add back the gems that resque-web required
that we had come to expect be part of our
package.


Gemfile
```diff
@@ -29,24 +29,19 @@ gem 'omniauth-shibboleth', '~> 1.2', '>= 1.2.1'
...
 gem 'rubyzip'
 gem 'sanitize'
 # Use SCSS for stylesheets
 gem 'sass-rails', '~> 5.0'
-# Need to constrain the version of Sinatra for resque-web, see https://github.com/sinatra/sinatra/issues/1055
 gem 'sidekiq'
-gem 'sinatra', '2.0.0.rc2'
 # Use sqlite3 as the database for Active Record
 gem 'sqlite3'
 gem 'therubyracer'
```
jenlindner: Can you remove this? I believe node.js is part of our default server build which obviates the need for this gem.

PR #883, opened 2018-02-21
added 'No Committee Members' field to the form

alexBLR:


alexBLR:
Here is a detailed request from Fran: https://github.com/emory-libraries/etd-issues/issues/14
I picked with adding an additional item to a dropdown list.

alexBLR:
@jenlindner, thank you for your revisions. I want to make sure I understand want you mean by disabled. Currently, committee member fields are disabled in both edit and create pages when "No Committee Members" are selected.
screen shot 2018-02-23 at 11 49 37 am 

Also, "add another committee member" button is disabled on create or when "No committee members" is selected.

Currently, the edit state committee members is kept in check with this snippet from etd_forms.rb
https://github.com/curationexperts/laevigata/blob/bfccc672c3a4434bf4075e6ac55eff116cc2c7df/app/forms/hyrax/etd_form.rb#L91-L92
And it seems to work when I create a new etd with no committee members and access the edit page it afterward. The state seems to be preserved with "No committee members" as well as disabled fields.

It preserves disabled state on edit page load based on this conditional that I added to the required attribute: https://github.com/curationexperts/laevigata/blob/bfccc672c3a4434bf4075e6ac55eff116cc2c7df/app/views/records/edit_fields/_committee_members.html.erb#L17

jenlindner:
I shouldn't have used the word disabled without clarifying. I mean making sure they don't contain values that show up in the submission review ('N/A'). 

I think the best approach for this would be to follow the precedent we use for opting out in the Supplemental Files and Embargo tabs. There's a well on the page with a checkbox that the user checks to indicate they don't want to submit data for this requirement, and when they do that the fields are disabled and do not contain any values. And when this is checked, no Committee Member row is created in the preview the user generates on the review tab, either. 

The validation also appears to be not quite working. To the user, it looks like I can submit an Etd with an empty value for a Non-Emory Committee Member because I get the green check and can generate a preview in the review, but nothing happens when I click submit. If I fix that by going back to the About Me tab and selecting 'No Committee Members' and submit, and then edit, the form loads in an invalid state, which should never happen. 

The create form with an empty value for a Non-Emory committee member:
create_form_on_qa

The review preview, which contains a row for a committee member that doesn't exist, with an empty space for their name and the default 'Emory' for their affiliation:

create_preview_on_qa

The edit loads with a red check for 'About Me':

edit_form_on_qa

These are all on QA.

And the edit case I was referring to earlier happens when you create an Etd without committee members, edit it to have one, and then edit again. Which is an edge case, but does mean there isn't enough logic to prevent it from happening.

bess:
@alexBLR Did you want to try to finish this in time for our next release?

alexBLR:
@bess, I have tests passing. And I am pretty much done with the story. Sorry for the delay, I had completely start over.

bess:
@alexBLR Okay, in that case do you want to close this PR and open a new one?

alexBLR:
Sounds good.

PR #898, opened 2018-02-27
Enforce embargoes whether they are strings or booleans

### LINKED ISSUE #897: Embargos not working
**EXPECTED BEHAVIOR**
The following ETDs have embargo terms that are not being reflected. 
- [x] https://etd.library.emory.edu/concern/etds/m039k489b?locale=en - Supposed to have a 1-year embargo on the Files.
- [x] https://etd.library.emory.edu/concern/etds/sb3978261?locale=en – Supposed to have a 2-year embargo on the Files
- [x] https://etd.library.emory.edu/concern/etds/vh53wv76t?locale=en – Supposed to have a 1-year embargo on Files, Table of Contents, and Abstract
- [x] https://etd.library.emory.edu/concern/etds/gm80hv35d - Supposed to have a 2-year embargo on Files, Table of Contents, and Abstract

**CURRENT BEHAVIOR**
Lisa Macklin reported in Slack (2/27) that the embargoes are not working on the above ETDs.

bess:
We have three fields that we use to determine what has been embargoed. Unfortunately, the application is not consistent about whether these should be a string value of "true" or a boolean value of `true`. This PR ensures that the presenter can interpret either case.

Fixes #897

### DIFF
--- a/app/models/solr_document.rb
+++ b/app/models/solr_document.rb
@@ -29,16 +29,31 @@ class SolrDocument
     self[Solrizer.solr_name('abstract')]
   end
 
+  # If this is a boolean, return the boolean value
+  # If this is a string, transform it into a boolean
+  # If the value is nil or can't be determined, assume it is embargoed
   def abstract_embargoed
-    self['abstract_embargoed_bsi']
+    return self['abstract_embargoed_bsi'] if self['abstract_embargoed_bsi']
+    return self['abstract_embargoed_tesim'].first.to_s == "true" if self['abstract_embargoed_tesim']
+    true
   end
 
+  # If this is a boolean, return the boolean value
+  # If this is a string, transform it into a boolean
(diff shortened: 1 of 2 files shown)


mark-dce:
@bess before we do a major upgrade of this code, I'd like to have a team discussion of whether we could have handled this more simply by extending the normal access policies:
* we have "restricted" as an existing policy with the behavior of being fully hidden from discovery
* we could add "restrict_files" to denote works that have only files embargoed but allow discovery including abstract & table of contents
* we could add "restrict_files_toc" to additionally restrict non-admin access to both files and table of  contents
* we could add "restrict_files_toc_abstract" as a policy to indicate items that should have core metadata discoverable, but should suppress all fields which might contain 'sensitive' information

This would allow us to use the out-of-box embargo functionality in Hyrax with minimal modification beyond allowing additional access policies.  There is a similar solution in place for Alexandria that could potentially be used as a model.

PR #903, opened 2018-03-01
Align Rollins programs with admin sets

### LINKED ISSUE #901: Still not mapping to admin sets correctly all the time
We used to have 1:1 correspondence between the menu options for School and Department and the admin sets they should go into. At some point this broke, and we keep thinking it's fixed and it isn't. 

- [ ] Write a test to systematically ensure that every combination of school / department / subfield maps to an admin set. Read the values directly from the menu file, so if the menu file ever changes and breaks this again it raises a red flag.

## Backtrace

    line 67 of [PROJECT_ROOT]/app/models/etd.rb: determine_admin_set
    line 73 of [PROJECT_ROOT]/app/models/etd.rb: assign_admin_set
    line 25 of [PROJECT_ROOT]/app/actors/hyrax/default_admin_set_actor.rb: ensure_admin_set_attribute!

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54218/faults/36903357)

bess:
Add tests so we know we have an admin set
for each option in the Rollins drop down menus.

Ultimately, we need to check our admin set config
info version control and really test against it
so things can't get out of sync, but this should
help in the meantime.

Fixes #901

### DIFF
--- a/spec/models/etd_spec.rb
+++ b/spec/models/etd_spec.rb
@@ -52,6 +52,32 @@ RSpec.describe Etd do
       etd.assign_admin_set
       expect(etd.admin_set.title).to contain_exactly("Epidemiology")
     end
+    # NOTE: These tests are here so that no one can change the name
+    # of a program or subfield in the authorities file and accidentally
+    # break the assignment of admin sets after ETD submission. If we change
+    # anything about the program names, or add a new one, we must ensure
+    # a corresponding admin set exists where that content will be routed.
+    it "can assign an admin set for every Rollins department except Environmental Health" do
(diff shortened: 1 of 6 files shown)


spec/models/etd_spec.rb
```diff
@@ -52,6 +52,32 @@
       etd.assign_admin_set
       expect(etd.admin_set.title).to contain_exactly("Epidemiology")
     end
+    # NOTE: These tests are here so that no one can change the name
```
jenlindner: 💯

PR #916, opened 2018-03-13
Committee members optional

alexBLR:


spec/fixtures/virus_checking/virus_check.txt
```diff
@@ -1 +1 @@
-X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*
+X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*
```
jenlindner: This looks so much better, @alexBLR. I'm going to run it locally this afternoon and get back to you. One thing though, what's going on in here? @bess - it looks like there's no need for this file in the pr, is that right?

alexBLR: @jenlindner @bess I always have trouble with virus_check.txt. My McAfee deletes it every time I run tests. I tried to disable it by specifying that specific path but it seems not to work.

bess: It might get deleted, that's okay. Just do `git checkout spec/fixtures/virus_checking/virus_check.txt` before you do your git commit. Or you could selectively git commit only the changes you intend to make, and not the removal of that file. But we do need the file to be in place as it's important for our virus checking tests. Would you please fix this and amend your commit?

alexBLR: @bess but the file is still present I just included that change as a part of the commit.

bess: @alexBLR You aren't changing anything about virus_check in this feature, right? So there should not be anything about virus check in your PR. Do you need help amending a PR? Would it be easier to open a new one?

alexBLR:
@bess @jenlindner I got rid of the virus_check from the PR.

alexBLR:
@jenlindner I updated the pull request with requested changes.

jenlindner:
@alexBLR Looks good! I've got the branch running locally and need to check out the back end and edit a little more, will let you know.

tmiles2:
@alexBLR @jenlindner can we add this work to the 4/10 release

jenlindner:
@alexBLR @tmiles2 I think so, I'll just check with some other folks on the DCE side in case we want to test it on QA, which has been devoted to ingest testing, but should be available to us shortly.

jenlindner:
@tmiles2 and @alexBLR, sorry, it took a little tweaking to get the branch onto QA. It's there now, and we're just giving it a run-through.

tmiles2:
@jenlindner  Thanks  @fpici keeping you in the loop

tmiles2:
@jenlindner I closed this by accident sorry, can you fix that

jenlindner:
@alexBLR and @tmiles2 This looks good, everyone approved. Merging now 💯

PR #945, opened 2018-03-30
Add read-only mode

### LINKED ISSUE #944: Add a read only flipper
- [ ] Add a flipper so the system can be put in read-only mode

bess:
Add a FlipFlop feature toggle that puts
the system into read-only mode.
This will allow us to more easily migrate
data behind the scenes.

Closes #944

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,9 +7,26 @@ class ApplicationController < ActionController::Base
   # Adds Hyrax behaviors into the application controller
   include Hyrax::Controller
   include Hyrax::ThemedLayoutController
+
+  # Check to see if we're in read_only mode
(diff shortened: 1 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -7,10 +7,27 @@ class ApplicationController < ActionController::Base
...
   with_themed_layout '1_column'
 
   protect_from_forgery with: :exception
 
+  # What to do if read_only mode has been enabled, via FlipFlop
+  # If read_only is enabled, redirect any requests that would allow
+  # changes to the system. This is to enable easier migrations.
+  def check_read_only
+    return unless Flipflop.read_only?
+    # Exempt the FlipFlop controller itself from read_only mode, so it can be turned off
```
no-reply: :+1:

Issue #955, opened 2018-04-02
Nagios monitoring for sidekiq


bess:
Add nagios monitoring for our sidekiq jobs. Ensure they are running and that the queues aren't getting backed up.

PR #956, opened 2018-04-02
Add okcomputer check for sidekiq latency

bess:
Connected to #955

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -26,3 +26,11 @@ end
 
 OkComputer::Registry.register "etd_load", EtdLoadCheck.new
 OkComputer::Registry.register "smtp",     SmtpCheck.new
+
+# Measure the latency of a sidekiq queue against some acceptable threshold measured in seconds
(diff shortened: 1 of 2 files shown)


config/initializers/okcomputer.rb
```diff
@@ -26,3 +26,9 @@ def check
 
 OkComputer::Registry.register "etd_load", EtdLoadCheck.new
 OkComputer::Registry.register "smtp",     SmtpCheck.new
+
+# Measure the latency of a sidekiq queue against some acceptable threshold measured in seconds
+# Latency == the difference between when the oldest job was pushed onto the queue versus the current time.
+# In other words, we want ingest jobs to run in < 5 minutes, and derivatives jobs to run in < 15 minutes.
+OkComputer::Registry.register 'ingest_queue', OkComputer::SidekiqQueueLatencyCheck.new(queue: :ingest, threshold: 300)
```
no-reply: I think we need a check for `:default` as well. Some jobs (`ProquestJob`, `GraduationJob`, and perhaps one or two others) are still queuing there.

Issue #924, opened 2018-03-16
On staging - Submitting to any Rollins Environmental Health subfield causes an error


shanitajohn:
**ISSUE** 
Currently, on staging-etd, attempting to submit an ETD from any of the Rollins, Environmental Health subfield results in the error message, "The page you were looking for doesn’t exist. You may have mistyped the address or the page may have moved." 

**ACCEPTANCE CRITERIA** 
- [ ] I should be able to submit an ETD via the Rollins School of Public Health, Environment Health Department, using subfield - Environmental Health - MPH 
- [ ]  I should be able to submit an ETD via the Rollins School of Public Health, Environment Health Department, using subfield - Environmental Health Epidemiology - MPH & MSPH
- [ ] I should be able to submit an ETD via the Rollins School of Public Health, Environment Health Department, using subfield - Global Environmental Health  - MPH

RachelDCE:
This is still broken in v1.5.1

RachelDCE:
I was able to submit successfully with all three subfields.

PR #958, opened 2018-04-03
Use a factory method for department and subfield services

no-reply:
Hiding this behavior in the `EtdHelper` makes it hard to test. Using a factory
method allows us to easily check that a given `school` or `department` yields a
correct service result.

For now, the factory just uses a case statement. We could switch to a registry
or some other pattern for matching schools or departments to their respective
services.

Several services had been renamed, but not corrected in bcb83332. These are
fixed here and regression tests are added.

Connected to #924.


no-reply:
superseded by #962

PR #959, opened 2018-04-04
Toggle read-only mode with a rake task

bess:
This lets our nightly backup process put the
system into read-only mode so it can stay on
while we make backups.


bess:
@no-reply My goal is to enable no-downtime backups. Feedback welcome on whether this is missing anything or could be better structured.

PR #962, opened 2018-04-04
Rollins and frontend fixes

jenlindner:
Needs review, particularly for accuracy and internal alignment of department and subfield names, and needs thorough user acceptance testing for school, department and subfield validation. One known issue is storing department names in the js, which creates an extra maintenance burden and a point of potential failure. Some of the feature tests could probably benefit from filling fields out of the sequential order they appear in on the page, additionally, there is a 5-second sleep added to the edit spec solving a timing issue that could possibly be better addressed. Refactoring the front end validation, which has grown complicated due to multiple dynamic and interdependent interactions, might be a good idea.

### DIFF
--- a/spec/features/edit_etd_spec.rb
+++ b/spec/features/edit_etd_spec.rb
@@ -132,7 +132,6 @@ RSpec.feature 'Edit an existing ETD', :perform_jobs do
       let(:dept) { 'Biological and Biomedical Sciences' }
       let(:subfield) { ['Genetics and Molecular Biology'] }
       let(:attach_supp_files) { true }
-
       let(:embargo_attrs) do
         {
           files_embargoed: true,
@@ -145,7 +144,7 @@ RSpec.feature 'Edit an existing ETD', :perform_jobs do
       scenario "edit a field", js: true do
         visit hyrax_etd_path(etd)
         click_on('Edit')
-
+        sleep(5)
         # Verify existing data in About Me tab
         expect(find_field('Student Name').value).to eq attrs[:creator].first
         expect(find_field('Student Name')).not_to be_disabled
@@ -157,6 +156,7 @@ RSpec.feature 'Edit an existing ETD', :perform_jobs do
         expect(find_field('School')).not_to be_disabled
         expect(find_field('Department').value).to eq attrs[:department].first
         expect(find_field('Department')).not_to be_disabled
+
         expect(find_field('Sub Field').value).to eq attrs[:subfield].first
         expect(find_field('Sub Field')).not_to be_disabled
         expect(find_field('Degree').value).to eq attrs[:degree].first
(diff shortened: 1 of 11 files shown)


app/controllers/hyrax/etds_controller.rb
```diff
@@ -41,6 +42,12 @@ def document_not_found!
       raise CanCan::AccessDenied.new(nil, :show)
     end
 
+    # we need to pass a nil to Hyrax when we remove the subfield,
+    # in order for Hyrax to remove and persist the etd correctly
+    def nil_to_delete_subfield(params)
```
no-reply: Should we move this to the `EtdForm`? There's already data transformation going on there, and it seems more robust to get this kind of transformation out of the MVC loop. See: https://github.com/curationexperts/laevigata/blob/b2f35b03a26e8ab86f9cfdc629941d993ced4f17/app/forms/hyrax/etd_form.rb#L150-L176

spec/features/edit_etd_spec.rb
```diff
@@ -145,7 +144,7 @@
       scenario "edit a field", js: true do
         visit hyrax_etd_path(etd)
         click_on('Edit')
-
+        sleep(10)
```
no-reply: Can you comment on why this sleep is now so long?

no-reply:
This looks okay to me. I'd like to refactor to push the data transformation out of the controller.

Also, could you rebase to remove the duplicated commits and get a commit message on the new changes? Particularly, that message should address the addition of the long sleep. The need for that is otherwise opaque to me.

spec/fixtures/vcr_cassettes/box_error.yml
```diff
@@ -1412,4 +1412,440 @@ http_interactions:
                 ns005:hasMember        <http://127.0.0.1:8986/rest/test/zw/12/z5/28/zw12z528p> .
     http_version: 
   recorded_at: Fri, 16 Mar 2018 18:13:06 GMT
+- request:
```
no-reply: Is this VCR cassette necessary?

app/controllers/hyrax/etds_controller.rb
jenlindner: Yes, happy to. We're also doing some transformations in the controller but no need for more.

spec/features/edit_etd_spec.rb
jenlindner: It could be less, I'll test locally to find whether 5 works, I just erred on the side of safer. We might want to consider this at some point too: https://github.com/jeremyf/capybara-maleficent.

spec/fixtures/vcr_cassettes/box_error.yml
jenlindner: Probably not, I was surprised to see it, but didn't look into when it appeared or why. I'm happy to rebase and squash the commits too, and edit the commit message.

PR #963, opened 2018-04-04
Adds URLs to ETD show page, and degree awarded notification.

rotated8:
No tests, as these are in views, although I'm happy to add them.

Here's a before:
before

And after:
after

### DIFF
--- a/app/views/hyrax/base/show.html.erb
+++ b/app/views/hyrax/base/show.html.erb
@@ -5,7 +5,9 @@
       <%= render 'work_title', presenter: @presenter %>
     </header>
 
+    <span>Permanent URL: <%= main_app.hyrax_etd_url(@presenter.id) %></span>
     <% unless @presenter.workflow.state_label == 'Approved' %>
+      <br />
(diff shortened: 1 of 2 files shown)


app/views/hyrax/base/show.html.erb
```diff
@@ -5,7 +5,9 @@
       <%= render 'work_title', presenter: @presenter %>
     </header>
 
+    <span>Permanent URL: <%= Rails.application.routes.url_helpers.send(@presenter.model_name.singular_route_key + '_url', @presenter.id) %></span>
```
no-reply: Why do we need to do `#send` here? Don't we have access to the `url_helpers` natively in the view?

rotated8: That was the idiom I saw in the code base (https://github.com/curationexperts/laevigata/blob/master/app/services/hyrax/workflow/laevigata_notification.rb#L21). `Rails.application.routes.url_helpers.hyrax_etd_url(@presenter)` also works, but `hyrax_etd_url(@presenter)`, and `url_for(@presenter)` do not.

rotated8:
`main_app.hyrax_etd_url(@presenter.id)` works. PR updated.

PR #967, opened 2018-04-06
Fallback admin set

no-reply:
The unassigned `AdminSet` is introduced as a catch-all for submitted ETDs that
couldn't be automatically assigned to another AdminSet. It has a generic
one-step approval process.

The purpose of this change is to avoid raising user facing errors when complex
internal admin set logic fails. Ideally, no items would ever be assigned to this
set; when they are, however, they can be triaged and repaired by
administrators. This avoids letting issues impact students at submission time.

Related to #924.

### DIFF
--- a/app/models/etd.rb
+++ b/app/models/etd.rb
@@ -6,6 +6,8 @@ class Etd < ActiveFedora::Base
   include ::Hyrax::BasicMetadata
   include ::ProquestBehaviors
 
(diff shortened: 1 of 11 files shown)


no-reply:
I haven't checked how easily an item put into "Unassigned admin set" can be moved to another one, or what impact this has on workflow or notifications.

PR #978, opened 2018-04-09
Presenter permission badge

jenlindner:
@no-reply this could maybe use a test for the new MemberPresenterFactory class, and a check for the missing badge in a feature spec - ?


app/presenters/etd_presenter.rb
```diff
@@ -23,6 +23,14 @@ class EtdPresenter < Hyrax::WorkShowPresenter
            :copyright_question_three,
            to: :solr_document
 
+  # we want to override the permission_badge method the FileSetPresenter class.
+  # In our Etds, the permission_badge method should not return html that indicates visibility settings (we handle embargos differently than Hyrax).
+  # We pass our own EtdFileSetPresenter, inside this factory, with the method that returns a benign empty string, instead.
+
+  def member_presenter_factory
+    Hyrax::EtdMemberPresenterFactory.new(solr_document, current_ability, request)
```
no-reply: I'm seeing: `uninitialized constant Hyrax::EtdMemberPresenterFactory`. Where is this defined?

Also:
> In our Etds, the permission_badge method should not return html that indicates visibility settings (we handle embargos differently than Hyrax).

Is this true? My understanding is that the `Etd` badges are already correct. It's only the `FileSet` views that we want to change, right?

app/presenters/file_set_presenter.rb
```diff
@@ -12,4 +12,8 @@ def supplementary?
     return true if solr_document["pcdm_use_tesim"] == "supplementary"
     false
   end
+
+  def permission_badge
```
no-reply: :+1:

app/views/hyrax/base/_primary_pdf.html.erb
```diff
@@ -9,7 +9,8 @@
         <th><%= t('.actions') %></th>
       </tr>
     </thead>
-    <% if member = @presenter.member_presenters.select { |mp| mp.solr_document["pcdm_use_tesim"].include?("primary")}.first %>
+    <% if member = @presenter.member_presenters.select { |mp| mp.primary? }.first
```
no-reply: Is this related? Can it be a different PR?

app/controllers/hyrax/file_sets_controller.rb
```diff
@@ -0,0 +1,6 @@
+module Hyrax
+  class FileSetsController < ApplicationController
+    include Hyrax::FileSetsControllerBehavior
+    self.show_presenter = EtdFileSetPresenter
```
no-reply: Where does `EtdFileSetPresenter` come from? I only see `FileSetPresenter`, which seems to be the one we want.

app/presenters/etd_file_set_presenter.rb
```diff
@@ -0,0 +1,17 @@
+class EtdFileSetPresenter < Hyrax::FileSetPresenter
```
no-reply: Do we need a new presenter? This one seems identical to https://github.com/curationexperts/laevigata/pull/978/files#diff-b0e29fd341e629e16c71110ec9ee28fa, yes?

jenlindner: We do. It contains the permission_badge method and refactored primary and supplementary methods. I pushed up the files last night in a bit of a messy state, sorry. We do not need the identical FileSetPresenter, however, deleted below.

app/views/hyrax/base/_primary_pdf.html.erb
jenlindner: It could be. It's related in that we are refactoring the presenter and making sure it gets used, and that these methods specifically are used in views as well, in this commit. The main reason we wanted our own fileset presenter was to distinguish between primary and supplementary filesets, so this commit makes sure it does its main job. We wanted to remove the badge as well, but as long as the commit message is clear about these things, I'm not sure we need two separate PRs. But if you'd prefer them that's fine with me.

spec/features/show_etd_spec.rb
```diff
@@ -103,6 +103,18 @@
     logout
   end
```
jenlindner: These tests confirm the behavior we are adding (and confirm the need for the EtdMemberPresenterFactory, which assigns our filesetpresenter to its file_presenter_class property).

Issue #985, opened 2018-04-13
Log the object id in all errors


bess:
- [ ] I should be able to search the error log to find everything associated with a given object

PR #994, opened 2018-04-17
Catch any places where we can't update post-graduation email

bess:
Make these errors more actionable.
Often they are caused by fake (testing) data, but we can't
just throw them away in case there's ever a real problem.
This should make it obvious whether it's a real problem for
a real student.

Connected to #985


no-reply:
Do we want to log and swallow the error, or log and reraise? Is the intent for the job to not retry?

bess:
@no-reply That's right, we do NOT want to re-run the job just because the email can't be assigned.

PR #1066, opened 2018-06-05
enabling flipflop to allow for safe UI development

jenlindner:


no-reply:
Closing this per discussion on slack call.

Issue #1065, opened 2018-06-04
Employ a feature flipper to do front end work behind


jenlindner:

jenlindner:
**ACCEPTANCE CRITERIA**
- [ ] uses FlipFlop to control check for new UI
- [ ] a new, separate controller with #new action is only access point to new UI view
- [ ] FlipFlop employed only in routes file to control access

PR #1077, opened 2018-06-11
institute a flag for new UI development

jenlindner:
This commit creates a flag which determines whether the work creating a new UI and partial-save feature is enabled or not. The flag is available via Rails.application.config_for('new_ui') and defaults to false. It is now employed in the routes file and our .travis.yml, in which we have a build for each flag value. Addresses #1065

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 dist: trusty
 language: ruby
-# cache: bundler
+
 sudo: false
 services:
   - redis-server
@@ -20,7 +20,11 @@ script:
   - bundle exec rubocop
   - xvfb-run -a bundle exec rake ci
 env:
-  - QMAKE=/usr/lib/x86_64-linux-gnu/qt5/bin/qmake
+  global:
+    - QMAKE=/usr/lib/x86_64-linux-gnu/qt5/bin/qmake
+  matrix:
+    - NEW_UI_ENABLED=false
+    - NEW_UI_ENABLED=true
 addons:
   apt:
     packages:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -16,11 +16,16 @@ rvm:
   - 2.4.2
 before_script:
   - bundle exec rake db:create
+  - bundle exec rake db:migrate
```
no-reply: Is this migration necessary with each build because of the schema changes behind the feature flag? Would it be okay to just run the migrations up-front and remove this moving part?

config/routes.rb
```diff
@@ -27,7 +27,17 @@
   mount Hyrax::Engine, at: '/'
   resources :welcome, only: 'index'
   root 'hyrax/homepage#index'
+  # While we work on new UI architecture, keep it accessible only when new_ui is true (see config/new_ui.yml).
+
+  if Rails.application.config_for('new_ui').fetch('enabled', false)
+    get '/concern/etds/new', to: 'in_progress_etds#new'
```
no-reply: Should this just direct to `get '/concern/etds/new', to: 'hyrax/etds#new'` temporarily? The second build is failing with a missing constant for the `:in_progress_etds` controller.

.travis.yml
no-reply: I guess for the moment there are no migrations. Can we just drop the line?

jenlindner: Yes, it's just a relic

config/routes.rb
jenlindner: Yep. One sec.

.travis.yml
jenlindner: Oh actually I got db errors working with the upgrade branch and put this in. I'm still taking it now but just so you know.

Issue #1072, opened 2018-06-06
Create a new submission form for in-progress-etds


jenlindner:
**ACCEPTANCE CRITERIA**
 - [x] Contains the 4 initial fields: name, email, graduation date and submission type

![1 about me](https://user-images.githubusercontent.com/3064318/41296753-ddd848b6-6e1a-11e8-922d-367b366272d4.png)

Issue #1067, opened 2018-06-05
Create a separate submission model with Name, E-Mail, Graduation, & Submission Type


jenlindner:
This is just the console accessible active record object. UI will be addressed in later tickets. So, this is a model with four string fields.

**ACCEPTANCE CRITERIA**
 - [x] an ETD Rails model with four (string) fields

PR #1078, opened 2018-06-11
mvc for new ui

jenlindner:
This commit builds a new controller and some very minimal views for a very simple InProgressEtd model, behind the new_ui_enabled flag. It adds an RSpec config to set this flag to false for feature tests. The new controller allows for find_or_create_by a user_ppid property, which addresses the need for partial saves and one in_progress_etd per student. It currently requires a db migration, which will be added in a subsequent PR. Addresses #1072 and #1067

### DIFF
--- a/app/controllers/in_progress_etds_controller.rb
+++ b/app/controllers/in_progress_etds_controller.rb
@@ -1,0 +1,35 @@
+class InProgressEtdsController < ApplicationController
+  def new
+    @in_progress_etd = InProgressEtd.find_or_create_by(user_ppid: current_user.id)
+  end
+
+  def create
+    @in_progress_etd = InProgressEtd.new(in_progress_etd_params)
+    @in_progress_etd.save
+    redirect_to @in_progress_etd
+  end
+
+  def edit
+    @in_progress_etd = InProgressEtd.find(params[:id])
+  end
+
+  def update
+    @in_progress_etd = InProgressEtd.find(params[:id])
+
+    if @in_progress_etd.update(in_progress_etd_params)
+      redirect_to @in_progress_etd
+    else
+      render 'edit'
+    end
+  end
+
+  def show
+    @in_progress_etd = InProgressEtd.find(params[:id])
+  end
+
+  private
+
+    def in_progress_etd_params
+      params.require(:in_progress_etd).permit(:name, :email, :graduation_date, :submission_type)
+    end
+end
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -16,6 +16,7 @@ rvm:
   - 2.4.2
 before_script:
   - bundle exec rake db:create
+  - bundle exec rake db:migrate
```
no-reply: Can we drop this back out?

app/controllers/in_progress_etds_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def new
+    @in_progress_etd = InProgressEtd.find_or_create_by(user_ppid: current_user.id)
+  end
+
+  def create
+    @in_progress_etd = InProgressEtd.new(in_progress_etd_params)
+    if @in_progress_etd.save
+      redirect_to @in_progress_etd
+    else
+      render 'new'
```
no-reply: Do we ever hit this branch? Only if `#save` fails because `@in_progress_etd` is invalid?

spec/rails_helper.rb
```diff
@@ -126,6 +126,10 @@
   #   describe "problem test", verify_partial_doubles: false do
   #     ...
   #   end
+  config.before(:all, type: :view) do
+    skip("View tests run only when NEW_UI_ENABLED") if ENV["NEW_UI_ENABLED"] == 'false'
```
no-reply: :+1:

app/controllers/in_progress_etds_controller.rb
jenlindner: @no-reply Yes. Since we don't have the validations I'll take it out. Just following very boilerplate structures right now.

spec/rails_helper.rb
jenlindner: @no-reply Is this approved, then?

spec/controllers/in_progress_etds_controller_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  if Rails.application.config_for('new_ui').fetch('enabled', false)
+    describe 'POST create' do
+      it 'persists a new InProgressEtd' do
+        post :create, params: { in_progress_etd: { name: 'Maud Gonne' } }
+
+        expect(InProgressEtd.all.count).to eq 1
+      end
+
+      it 'persists a new InProgressEtd with all of its attributes' do
+        post :create, params: { in_progress_etd: { name: 'Maud Gonne', email: 'maud@sinnfein.org', graduation_date: '1916', submission_type: 'Honors Thesis' } }
```
no-reply: :joy: :ireland:

no-reply:
Coveralls issue seems to be due to the two parallel builds. @jenlindner indicated she is looking for a solution to let coveralls know what's up.

If we can't find one, this is maybe a topic for the team meeting.

Issue #1081, opened 2018-06-11
Resume submission when I click "Submit My ETD"


jenlindner:
**ACCEPTANCE CRITERIA**
- [x] I get started with any data I had already entered in a previous session.

Issue #1080, opened 2018-06-11
Edit my InProgressETD submission from the review form


jenlindner:
**ACCEPTANCE CRITERIA**
- [x] Add an edit button to the review page 
- [x] The button takes me to an edit form
- [x] I can resubmit from the edit form 
- [x] I see an updated review page

PR #1086, opened 2018-06-12
add table for in-progress-etds to rails db

jenlindner:
This commit adds a table to the Rails db, for persisting in-progress-etds. The table contains all of the ETD properties and the user's ppid. The addition of a user id in this table enables a very simple mechanism for enforcing one in-progress-etd per user, and editing with partial saves (find_or_create_by(:user_ppid) in InProgressEtds#new). Addresses #1080 and #1081.


jenlindner:
@no-reply yes, isn't that in the message above? "Addresses #1080 and #1081."

no-reply:
Thanks. I'm a bit slow this morning. :dancer:

Issue #1094, opened 2018-06-13
Support inactive authority fields


no-reply:
When editing an existing object:

  - [x] only active authority terms are displayed by default.
  - [ ] if the item includes an inactive term, that term is included as the selected option when building the form.
    - [x] department
    - [ ] subfield
    - [x] graduation dates
    - [ ] partnering agency
    - [x] Schools (hopefully no surprises here...)
    - [x] Degree
    - [ ] Submission Type
    - [ ] Research Fields
    - [x] Language

no-reply:
I can't seem to get waffle to work for this issue, but at least it's not closed and in progress anymore!

no-reply:
This work is avoided by the new form, which does not use Hyrax's (or Laevigata's custom) authority services. It needs to be redone in light of that.

jenlindner:
@no-reply graduation dates and research fields are the only authorities that currently have any terms with active: false.

PR #1095, opened 2018-06-13
Support inactive QA terms

no-reply:
Add a helper that can be used when creating form select fields from QA authorities. This is implemented upstream in samvera/hyrax#3122, so a guard is included to flag removal upon a future upgrade.

Use the helper when building form fields; change the existing forms to use active terms only.

Connected to #1094.

### DIFF
--- a/spec/services/hyrax/laevigata_authority_service_spec.rb
+++ b/spec/services/hyrax/laevigata_authority_service_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe Hyrax::LaevigataAuthorityService do
+  subject(:select_service) { described_class.new('graduation_dates') }
+
   describe '.for' do
     it 'raises an error if no arguments are passed' do
       expect { described_class.for }.to raise_error ArgumentError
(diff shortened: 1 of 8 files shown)


no-reply:
@jenlindner I'd like to rebase this on #1096 and see if this error pops up again.

I can't see immediately how those JS errors would be due to changes in this PR, or how they would be happening intermittently. I wonder if it's just a hiccup in the travis env?

In any case, I've attached the log here, and I guess we'll see if this pops up again.
[log.txt](https://github.com/curationexperts/laevigata/files/2099924/log.txt)

jenlindner:
@no-reply -- sounds good

spec/services/hyrax/laevigata_authority_service_spec.rb
```diff
@@ -51,4 +53,23 @@
...
+        .to eq [[['2009', '2009']], { class: 'moomin force-select' }]
+    end
+
+    it 'does not add an empty/missing value' do
+      expect(select_service.include_current_value('', :idx, render_opts, html_opts))
+        .to eq [render_opts, html_opts]
+    end
+
+    it 'does not add an active current value' do
+      expect(select_service.include_current_value('Spring 2019', :idx, render_opts.dup, html_opts.dup))
```
val99erie: Are you missing expected value here?

no-reply: looks like!

PR #1098, opened 2018-06-15
New UI test filtering

jenlindner:
This commit adds test configuration that ensures new_ui view and feature tests only run when the new_ui flag is true, and vice-versa. It also adds some feature tests covering the find-or-create-by behavior, as a result of which users can only work on one in-progress-etd until they've published it.


spec/rails_helper.rb
```diff
@@ -119,22 +119,36 @@
...
+  end
+
+  # because we need to negate with skip, these conditionals are a little awkward, but they do control the flow correctly.
+  # this is the flow we want:
+  # when env new_ui is true, only run tests tagged new_ui
+  # when env new_ui is false, only run tests without new_ui tag
+
+  config.before(:each, type: :feature) do |example|
+    if new_ui
+      skip('Hyrax Feature tests do not run when NEW_UI_ENABLED') unless example.metadata[:new_ui] == true
```
no-reply: `unless example.metadata[:new_ui]` instead?

spec/rails_helper.rb
```diff
@@ -119,22 +119,36 @@
...
+    end
+
+    skip("New UI Feature tests run only when NEW_UI_ENABLED") if new_ui == false && example.metadata[:new_ui] == true
```
no-reply: `if !new_ui && example.metadata[:new_ui]` ?

spec/rails_helper.rb
```diff
@@ -119,22 +119,36 @@
...
+  end
+
+  config.before(:all, type: :view) do
+    skip("View tests run only when NEW_UI_ENABLED") if Rails.application.config_for(:new_ui).fetch('enabled', false) == false
```
no-reply: `unless Rails.application.config_for(:new_ui).fetch('enabled', false)`?

Do we want to use the `new_ui` pattern here, too? Alternatively, can we remove it and just check `config_for(:new_ui)` in each place for feature tests?

jenlindner:
@no-reply 👍 how's this?

PR #1119, opened 2018-06-21
disable deprecation warnings in ci

jenlindner:
Because the slowness they seem to add seems to be interfering with our Travis builds.

### DIFF
--- a/config/initializers/hyrax.rb
+++ b/config/initializers/hyrax.rb
@@ -206,10 +206,12 @@ Hyrax::CurationConcern.class_eval do
     @work_middleware_stack ||= actor_factory.build(Hyrax::Actors::Terminator.new)
     return @work_middleware_stack if args.empty?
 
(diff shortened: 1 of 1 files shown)


mark-dce:
Before we do this, could you:
 1) provide a comparison of timings of the average of three runs with deprecation warnings and three runs without
2) run the test suite with the [--profile](https://relishapp.com/rspec/rspec-core/docs/configuration/profile-examples) option and evaluate whether there are any particularly slow tests that are redundant or could be refactored to run more quickly

If we have that information in hand, I'd be much more comfortable having a discussion about whether to turn off deprecation warnings or not.

mark-dce:
I'd also be willing to discuss devoting a 1-2 day spike to resolving deprecations if that was useful.

no-reply:
I'd like to just cleanup the deprecations today.

no-reply:
I started work on the deprecation cleanup at: https://github.com/curationexperts/laevigata/pull/1135

PR #1125, opened 2018-06-21
 Improve embargo management UX and remove "Deactivate Embargo" button 

### LINKED ISSUE #1108: Fix Lift Embargo
See Hyrax ticket https://github.com/samvera/hyrax/issues/3058

Consider using the term "Lift Embargo" instead of deactivate which seems ambiguous: deactivate=delete (keep at restricted level) OR deactivate=lift(set to less restrictive level)

Clearest would be buttons that say "Set to X now" where X = the end embargo state

no-reply:
The terminology in embargo management pages is frequently overly wordy and unclear out of the box. Particularly, "deactivate" is overloaded: meaning "release"/"lift" when the embargo release date is past, and meaning "reset item permissions to embargoed and mark the embargo inactive" otherwise. In
`list_expired_active_embargos`, the former is always the case, so we say "Release" to reduce confusion ("Release", rather than "lift", is the terminology used elsewhere).

While we're here, we tighten up some other values to avoid overly wordy and vague text, (e.g. "Cancel and manage all embargoes" becomes "Manage Embargoes").

The counter-intuitive "Deactivate Embargo" button is also removed.

Related to curationexperts/nurax#205 & samvera/hyrax#3058.

Closes #1108.

### DIFF
--- a/app/views/hyrax/embargoes/edit.html.erb
+++ b/app/views/hyrax/embargoes/edit.html.erb
@@ -1,0 +1,57 @@
+<% cc_type = curation_concern.human_readable_type %>
+
+<% provide :page_header do %>
+  <h1><%= t('.manage_embargoes_html', cc: curation_concern, cc_type: cc_type) %></h1>
+<% end %>
+
+<div class="panel panel-default tabs">
+  <div class="panel-heading">
+    <h2 class="panel-title"><%= t('.header.current') %></h2>
+  </div>
+  <div class="panel-body">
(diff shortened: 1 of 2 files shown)


mark-dce:
@no-reply I just approved and merged this pull request because it definitely helps resolve the immediate issue in Laevigata.  I'm wondering though, how we leave ourselves a breadcrumb to remember to go back and revisit the embargo form if we end up with something as good or better upstream in Hyrax.  Thoughts?

Issue #1029, opened 2018-05-16
Display School Configuration


mark-dce:
**STORY**
As a product owner or application manager, I would like a view that shows the configuration information for a school including;  
* Display Name
* Associated Departments
* Admin Set ID
* Departmental Approvers
* Departmental Subfields

**ACCEPTANCE**
- [x] I can access a screen that gives information similar to the mock-up below (there may be more info that should be included)
- [x] The screen is restricted to super-admin users
- [ ] [BONUS] I can dump this information to JSON or similar machine readable format that I could use in a manual or automated test script :)

**MOCK-UP**
![school configuration](https://user-images.githubusercontent.com/3064318/40148102-efb43ade-5931-11e8-8c8d-d6523fbcb1bc.png)

PR #1130, opened 2018-06-22
Story 1029: Display config for schools

val99erie:
The configured relationships between schools, departments, subfields,
admin sets, approving users, and workflows is complicated.  That
information is spread across a dozen yaml files and 5 or 10 database
tables.  It's really hard to understand the current configuration for a
running app.

This code attempts to make it easier to understand the config by
displaying the full config on a single web page for each school.

Some things to note:

* I made a SchoolsController to display the config info for each school,
which is only accessible by admin users.

* I pulled the logic for choosing the correct AdminSet into a new class
AdminSetChooser, so that I could share that logic between the Etd class
and the School and Department classes.

* I copied app/views/hyrax/dashboard/sidebar/_configuration.html.erb over
from hyrax so that I could insert the nav link for the schools config
page.

Connected to #1029

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -70,6 +70,11 @@ Metrics/MethodLength:
 Metrics/ModuleLength:
   Enabled: false
 
+Rails/Delegate:
+  Exclude:
+    - 'app/models/schools/department.rb'
+    - 'app/models/schools/subfield.rb'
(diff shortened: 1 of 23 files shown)


no-reply:
Connected to #1029

PR #1132, opened 2018-06-22
Update sprockets-rails

no-reply:
This is in response to critical security issue CVE-2018-3760.


no-reply:
This isn't necessary after the rails 5.1 upgrade.

PR #1134, opened 2018-06-22
Add webpacker and Vue

little9:
This commit adds webpacker and Vue to the codebase. Webpacker
is a way of managing front-end assets. Vue is a progressive
JS framework that is designed to be incrementally adoptable.

Additional applications need to be installed in the dev env
after this commit. You will need to install `yarn`, a JS package
manager similar to `gem`.

On a Mac you can install it like this:

`brew install yarn`

On other systems you can install it though `npm`:

`npm install -g yarn`

When running the dev env you will need to have the
`webpack-dev-server` running in another tab:

`bin/webpack-dev-server`


val99erie:
@little9 The notes you have in the comment about installing yarn and running the webpack server seem really useful.  Could you add them to the developer setup in the README?

little9:
@val99erie Good idea!

PR #1139, opened 2018-06-25
Changes ETD help language and link to ETD help

devanshu-m:
I am not sure why the WorkFlowSetup tests are failing. I have not made any changes to that piece. All tests are passing on my local machine.


no-reply:
This looks good.

Does this need to be rolled out on a particular timeline? `master` is currently the `laevigata` 2.0 (Hyrax 2.1) line, and we may need a backport to `1.6-stable` if a release is needed with the updated text.

cc: @bess

PR #1154, opened 2018-06-26
Yarn rake task fix

little9:
There is a bug in the current version of the webpacker gem that 
interferes with our capistrano deployment.

See:
https://github.com/rails/webpacker/issues/810

This adds a rake task which overrides the one provided by the gem so that yarn is called from the system version and not a binstub.

### DIFF
--- a/lib/tasks/yarn.rake
+++ b/lib/tasks/yarn.rake
@@ -1,0 +1,6 @@
+namespace :yarn do
+  desc "Install all JavaScript dependencies as specified via Yarn"
+  task :install do
+    system("yarn install --no-progress --production")
+  end
+end


lib/tasks/yarn.rake
```diff
@@ -0,0 +1,6 @@
+namespace :yarn do
+    desc "Install all JavaScript dependencies as specified via Yarn"
+      task :install do
+      system("yarn install --no-progress --production")
+    end
+end
```
no-reply: ending newline?

lib/tasks/yarn.rake
```diff
@@ -0,0 +1,6 @@
+namespace :yarn do
+    desc "Install all JavaScript dependencies as specified via Yarn"
+      task :install do
+      system("yarn install --no-progress --production")
```
no-reply: @bess do you have thoughts about this? I don't think the app is doing system installs elsewhere, and wanted to check if this is a pattern we're comfortable with.

lib/tasks/yarn.rake
little9: @no-reply Is this something that rubocop can check for?

mark-dce:
Coverage is only down in NEW_UI_ENABLED=false runs - which is expected

PR #1170, opened 2018-06-28
Add department dropdown to form

little9:
This commit adds the department dropdown
back to the "About Me" portion of the form.

Right now they are both on the "About Me" tab,
but the workflow diagrams have the department
dropdown on the next tab.

This changes the name of the `formSchema` object
to `formStore` to better reflect what that object
is.  The object is a store from the Flux pattern:
https://vuejs.org/v2/guide/state-management.html#Simple-State-Management-from-Scratch

That means that any changes to the state of the `formStore` object
should be done using methods on the `formStore` object itself. That
way the object can be used in many different components, with all
of the changes happening in a known location.


val99erie:
Could you rebase this branch?  It seems to have commits from a different PR mixed into it.

little9:
@val99erie yeah I'm going to need to clean up this PR. There was a PR merged yesterday that broke travis.

mark-dce:
@little9 the reason for having the department on the next tab is that it's dependent on the school choice.  I think it gives us a cleaner user experience to minimize the amount of dynamic content necessary on each tab.  

With them both on the same tab, we have to address the following states:
1. What is the initial content of the department dropdown when the school isn't selected
1. If a user changes school selection after selecting a department, do we keep the department setting or reset it?

If you draw it on the next tab, you always know which school has been selected first and only display an existing department value from previous user input if it matches one of the valid options for the school.

little9:
@mark-dce we made an issue for changing the behavior so that it happens on multiple tabs earlier: 

https://github.com/curationexperts/laevigata/issues/1194

mark-dce:
awesome - thanks!

Issue #1181, opened 2018-06-28
Setup unit testing for Vue 


little9:
See: https://vuejs.org/v2/guide/unit-testing.html

These should be runnable by doing `yarn test` and the travis yml should be updated to run these tests.

PR #1211, opened 2018-06-29
Add testing and linting for scripts in app/javascript

little9:
This commit sets up tesing and linting for the code in
app/javascript. The tests can be run with  `yarn test` and
linting can be run with `yarn lint`.

Testing is done with http://jestjs.io/ and can be used with
modern JS modules, Vue Single File Components, or plain JS.
There is an inital
test in app/javascript/test/app.spec.js that can be used
as a reference. The assertions that be made are documented here:
http://jestjs.io/docs/en/expect

The travis build will run `yarn test`.

Linting is done with the standard js style: https://standardjs.com/

Travis is not running `yarn lint` yet. There will need to be some
clean up of the exisiting code before making that part of the build
process if we want to make it part of the build process.

Related to #1181

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,5 @@
 {
-  "presets": [
+  "presets": ["es2015",
     ["env", {
       "modules": false,
       "targets": {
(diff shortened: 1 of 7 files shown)


jenlindner:
I think we might benefit from a ticket for getting linter to pass and running in it the travis build, thoughts @little9?

PR #1212, opened 2018-06-29
Remove redundant css in hyrax.scss

mark-dce:
Only the last setting for each value gets read, the first is redundant.


mark-dce:
What!  I'm deleting one line of redundant code - how can coverage decrease 5% from that???

Issue #1204, opened 2018-06-28
View all collections link causes exeption


mark-dce:
**STEPS TO REPRODUCE**
1. Login as an administrative user
1. Navigate to the dashboard
1. Click on "Collections"
1. Click on the "All Collections" tab (which links to /dashboard/collections )

**EXPECTED RESULTS**
A page displaying 0 or more collections similar to the "your collections" tab.

**ACTUAL RESULTS"
"We're sorry, but something went wrong."  
see https://app.honeybadger.io/projects/56103/faults/38420981

PR #1214, opened 2018-06-29
Fixed error when click 'All Collections' link. Story #1204

val99erie:
There are 2 parts to this fix:

1. The hyrax view assumes that the description returned by the presenter
is an array (and attempts to call `.first` on it).  That will cause an
error in the case where description is nil.  So I made a change to the
SolrDocument#description to return empty array if description is nil to
avoid raising an error.

2. In the case where the Collection does have a description, the
presenter/SolrDocument was returning nil.  This is because the
Collection model was using an indexer that doesn't index the
BasicMetadata.  (We missed a new indexer during our recent hyrax gem
upgrades.)

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -3,4 +3,5 @@ class Collection < ActiveFedora::Base
   include ::Hyrax::CollectionBehavior
   # You can replace these metadata if they're not suitable
   include Hyrax::BasicMetadata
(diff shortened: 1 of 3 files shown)


no-reply:
There is a small hyrax-side fix for this as well: https://github.com/samvera/hyrax/pull/3137

Issue #1162, opened 2018-06-27
School Dashboard should retain toolbar at left


mark-dce:
**ISSUE**
Compare clicking on the "Review Submissions" or "Works" links in the left navigation bar to clicking on "Schools".  In the first two cases, the navigation bar remains available while viewing the other content.

**ACCEPTANCE**
- [x] The dashboard navigation bar is displayed to the left of the "Schools" configuration views

PR #1215, opened 2018-06-29
Use dashboard layout when viewing school views. Story #1162

val99erie:


no-reply:
The failing test should be fixed by #1226.

Issue #1189, opened 2018-06-28
Add a rich text editor for abstract and table of contents on the My Etd tab


little9:
Investigate alternatives to TinyMCE,  but use TinyMCE if necessary.

little9:
Maybe Quill? 

https://quilljs.com/

PR #1216, opened 2018-06-29
Add Quill for editing abstract and toc

little9:
This adds the Quill editor for rich text editing
of the abstract and table of contents. There
will need to be a follow up PR that sets the
toolbar options.

The text that is output by Quill is bound to a
hidden input field with the correct name for
the form submission.

https://quilljs.com/

Related to #1189

### DIFF
--- a/app/javascript/app.vue
+++ b/app/javascript/app.vue
@@ -13,6 +13,20 @@
             <div v-if="input.label === 'School'">
               <school></school>
             </div>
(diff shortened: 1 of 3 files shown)


little9:
@mark-dce Is there a ticket from the past that describes what toolbar options are required for the rich text box. Right now they are all enabled, which I'm sure we will want to change: 

screenshot from 2018-06-29 15-48-16

PR #1225, opened 2018-07-02
Add dynamic Submission Type dropdown

little9:
This commit adds a component that renders the
the submission types dropdown on the form. It
does an XHR request to the QA endpoint to accomplish
this.

There is a basic spec in place, but to fully test
the component, there will need to be an Axios mock
in place. The spec currently triggers a node warning.

### DIFF
--- a/app/javascript/SubmittingType.vue
+++ b/app/javascript/SubmittingType.vue
@@ -1,0 +1,32 @@
+<template>
+    <div>
+        <label>Submission Type</label>
+        <select name="etd[submitting_type]" class="form-control">
+            <option v-for="submittingType in submittingTypes" v-bind:value="submittingType.id" v-bind:key='submittingType.id' v-if="submittingType.active">
+                {{ submittingType.label }}
+            </option>
+        </select>
+    </div>
+</template>
+
+<script>
+import Axios from "axios"
+export default {
+  data() {
+    return {
+      submittingTypeEndpoint: "/authorities/terms/local/submitting_type",
+      submittingTypes: {}
+    }
+  },
+  methods: {
+    fetchData() {
+      Axios.get(this.submittingTypeEndpoint).then(response => {
+        this.submittingTypes = response.data
+      });
+    }
+  },
+  created() {
+    this.fetchData()
+  }
+}
+</script>
--- a/app/javascript/app.vue
+++ b/app/javascript/app.vue
@@ -27,6 +27,9 @@
                </quill-editor> 
                <input class="quill-hidden-field" :name="etdPrefix(index)" v-model="input.value" /> 
             </div>
+            <div v-if="input.label === 'Submitting Type'">
+              <submittingType></submittingType>
+            </div>
             <div v-else>
               <label>{{ input.label }}</label>
               <input class="form-control" :name="etdPrefix(index)" v-model="input.value">
@@ -49,6 +52,7 @@ import { quillEditor } from 'vue-quill-editor'
 import 'quill/dist/quill.core.css'
 import 'quill/dist/quill.snow.css'
 import 'quill/dist/quill.bubble.css'
+import SubmittingType from './SubmittingType'
 
 let token = document
   .querySelector('meta[name="csrf-token"]')
@@ -83,7 +87,8 @@ export default {
   },
   components: {
     school: School,
-    quillEditor: quillEditor
+    quillEditor: quillEditor,
+    submittingType: SubmittingType
   },
   methods: {
     etdPrefix(index) {
--- a/app/javascript/test/SubmissionType.spec.js
+++ b/app/javascript/test/SubmissionType.spec.js
@@ -1,0 +1,27 @@
+/* global describe */
+/* global it */
+/* global expect */
+import { shallowMount } from '@vue/test-utils'
+import SubmittingType from 'SubmittingType'
+
+describe('SubmittingType.vue', () => {
+  it('renders a form', () => {
+    const wrapper = shallowMount(SubmittingType, {
+    })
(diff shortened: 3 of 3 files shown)


app/javascript/SubmittingType.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+    <div>
+        <label>Submission Type</label>
+        <select name="etd[submitting_type]" class="form-control" id="degree">
+            <option v-for="submittingType in submittingTypes" v-bind:value="submittingType.id" v-bind:key='submittingType.id' v-if="submittingType.active">
```
mark-dce: @no-reply @little9 It seems gross that we have to push the selection of active terms all the way to the UI.  It seems like this is something QA should handle for us (but it doesn't look like it does: https://github.com/samvera/questioning_authority/blob/master/lib/qa/authorities/local/file_based_authority.rb)  Should be consider a PR to QA that implements something like an "/authorities/terms/local/[vocab]/active" endpoint that only returns currently active terms?

I don't want to hold this PR for that, just wanted to start the discussion if it's useful.  We can carry it on elsewhere if that's useful.

app/javascript/SubmittingType.vue
```diff
@@ -0,0 +1,32 @@
...
+      submittingTypes: {}
+    }
+  },
+  methods: {
+    fetchData() {
+      Axios.get(this.submittingTypeEndpoint).then(response => {
+        this.submittingTypes = response.data
+      });
+    }
+  },
```
mark-dce: @little9 this feels like a lot of overhead to get a local list of entries.  I.E. we're making an http call to have a different service read a file that has less than 10 entries.  Should we build out QA format compatible VUE scripts that just read the config files directly?

app/javascript/app.vue
```diff
@@ -45,7 +49,8 @@ export default {
     }
   },
   components: {
-    school: School
+    school: School,
+    submissionType: SubmittingType
```
mark-dce: `submissionType: SubmittingType` is a little opaque and I wasn't keeping the two of them separate in my head when I first read the code.  Would it make more sense to make them more distinct, maybe something like `submissionType: DepositTypeList`?  (`PaperType`, `ThesisOrDissertation`, `DepositTypeSelection`)

little9: Hmm originally I had named these both `submissionType` and `SubmissionType`, but noticed that the params were actually "Submitting Type". Whenever I was renaming them I must have missed this one. So I was actually trying to make them more similar 😄 

My worry is that with some separate name someone else might not able to find the equivalent on the backend?

mark-dce: Ok, I can also live with them being identical.  It was the really close but not the same that seemed challenging.  There's still the issue of capitalization being slightly different, but hopefully folks will get that once they start using VUE.

app/javascript/SubmittingType.vue
little9: I wouldn't worry about this seeming like a lot of overhead. There are other lists managed by QA that only have 3 items. The requests are going to be totally unnoticeable in actual use, when you use autocomplete in Hyrax it's make requests whenever you type for example. 

Doing it in this way will mean that changes to the vocabs cascade appropriately to all the forms.

app/javascript/SubmittingType.vue
no-reply: Extending QA with some kind of query mechanism for attributes other than 'term' seems reasonable, but unnecessary here.

Pushing query to browser side might be unfortunate, though (do we always expose the QA endpoint to the public?).

Apart from that consideration, this seems like it's just a (partial) javascript client for QA, and I'm :ok: with that.

app/javascript/SubmittingType.vue
no-reply: I'm in favor of continuing to use `qa`'s HTTP API as the integration point until and unless that becomes an issue.

no-reply: If we are worried about the browser-side interactions, we could consider ways to avoid them that still draw the authority list from `qa` at runtime. That's my understanding of how `qa` is designed, and I think we shouldn't fight it unless we are ready to re-engineer more broadly.

app/javascript/SubmittingType.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+    <div>
+        <label>Submission Type</label>
+        <select name="etd[submitting_type]" class="form-control" id="degree">
```
no-reply: I'm not sure I'm following the labels and attributes here. We have "Submission Type", `name="etd[submitting_type]"` `id="degree"`.

Is it possible to normalize slightly?

A `<label for="whatever">` is probably called for as well?

no-reply:
Looks okay to me. I have some qs about label and attributes on the input, but I gather that's already been discussed somewhat?

no-reply:
@little9 Looking at the last round of failing tests on this, I noticed `Unhandled promise rejection` warnings from node. Does your latest fix that, or did that not creep in on this PR?

little9:
@no-reply when I determine the best way to mock the ajax requests in the jest tests , those warnings will go away

no-reply:
@little9 that seems okay. Can you get that info into the commit log?

little9:
@no-reply let me know if that's ok

app/javascript/SubmittingType.vue
mark-dce: Agreed it's not a priority right now  I'm just observing that VUE on top of QA introduces an abstraction on top of an abstraction that adds to cognitive overhead I need to maintain to parse that we're building a select box here that presents three options.

little9:
@no-reply Can you merge this? Each one of the dropdowns is going to need to be rebased after this is merged.

PR #1228, opened 2018-07-02
Don't stub `ActiveJob::Base` in tests

no-reply:
We use the test adapter for `ActiveJob`, so stubbing `Base` makes things more complex for no particular benefit. We can count on the adapter to manage which jobs are run during the test suite.


no-reply:
This looks like it's genuinely failing one spec. I'll close and revisit.

Issue #1219, opened 2018-06-29
"ETD" capitalization


mark-dce:
**ISSUE**
In listings where I see the work type (which might only be in Embargoes), I see new works submitted with 2.x code listed as "Etd" whereas works submitted with 1.x code are listed as "ETD" (all caps)

**ACCEPTANCE**
- [x] New submissions show the type as "ETD" in any relevant display views

<img width="1161" alt="1219 - etd capitalization" src="https://user-images.githubusercontent.com/3064318/42117704-d57358f8-7bc3-11e8-8f24-ab25019e4a56.png">

no-reply:
oh, that's wild. This mus be stored verbatim in the db!

no-reply:
Note that #1230 doesn't fix this issue on existing migrated data in QA and staging envs until a reindex takes place.

PR #1230, opened 2018-07-02
Use i18n to give the ETD model a human readable label

bess:
The previous system for doing this, model.human_readable_name,
has been deprecated. This seems to be the new preferred system
for giving a model a human readable label.

Connected to https://github.com/curationexperts/laevigata/issues/1219

### DIFF
--- a/spec/models/etd_spec.rb
+++ b/spec/models/etd_spec.rb
@@ -3,6 +3,12 @@
 require 'rails_helper'
 
 RSpec.describe Etd do
+  context "human readable type" do
+    it "capitalizes ETD because it is an acronym" do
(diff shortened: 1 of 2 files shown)


spec/models/etd_spec.rb
```diff
@@ -3,6 +3,12 @@
 require 'rails_helper'
 
 RSpec.describe Etd do
+  context "human readable type" do
+    it "capitalizes ETD because it is an acronym" do
+      expect(I18n.t "activefedora.models.etd").to eq "ETD"
```
no-reply: rubocop prefers `expect(I18n.t("activefedora.models.etd"))` which seems a bit more clear.

PR #1235, opened 2018-07-03
New ui/add tab validation - About Me

jenlindner:
This might be easiest and simplest way to validate an InProgressEtd on a per-tab basis. The name of the tab is added to the data, and a custom Validator class uses a current_tab? to check the tab name and determine whether to run. If it runs, it checks the fields in a list stored in the HyraxEtd Form class, via a new method in the etd class (based on similar methods and lists for the current ETD tab contents). 

It might be a little repetitive, and there might be other flaws. It needs tests. So, I want to discuss.

### DIFF
--- a/app/controllers/in_progress_etds_controller.rb
+++ b/app/controllers/in_progress_etds_controller.rb
@@ -6,8 +6,11 @@ class InProgressEtdsController < ApplicationController
   def create
     @in_progress_etd = InProgressEtd.new
     @in_progress_etd.data = prepare_etd_data.to_json
-    @in_progress_etd.save
-    redirect_to @in_progress_etd
+    if @in_progress_etd.save
+      redirect_to @in_progress_etd
+    else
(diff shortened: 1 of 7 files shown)


val99erie:
The per-tab validation seems like a good way to do it.

My first instinct was that we should decouple model validation from form validation, so the student can see in the form if any fields are missing or incorrect, but that doesn't prevent them from saving the record.  (i.e. The student should be able to save an in-progress ETD, no matter how many fields are missing or incorrect, and the model-level validation should only happen when the student is ready for the final submit that will convert their in-progress draft to a real `Etd` record.)

However, the more I thought about it, I couldn't figure out a way to do it where the UI wouldn't be pretty confusing for the student.  I think your per-tab validation will make a cleaner and easier-to-understand UI than any alternative that I could think of.

The nice thing about having separate validator classes per tab is that you might be able to share the validators between the `InProgressEtd` and the `Etd` classes.

Issue #1183, opened 2018-06-28
Add help text from the mockups to the Vue form


little9:
There is help text for each of the tabs that should be copied from the mock-ups.

Issue #1182, opened 2018-06-28
Change styling of the form to look like tabs


little9:
The Vue form should look like the mockups.

PR #1246, opened 2018-07-04
Add help text and style the form tabs

little9:
This adds the help text and styles the from so that it has tabs. 

This is for #1182 and #1183

### DIFF
--- a/app/javascript/formStore.js
+++ b/app/javascript/formStore.js
@@ -33,8 +33,8 @@ export var formStore = {
     },
     my_advisor: {
       label: 'My Advisor',
-      help_text: `Please provide some details about the people who supervised your submission. 
-      If your committee chair, thesis advisor, or committee members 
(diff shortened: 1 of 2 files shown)


app/javascript/formStore.js
```diff
@@ -87,7 +89,10 @@ export var formStore = {
     },
     my_files: {
       label: 'My Files',
-      help_text: '',
+      help_text: `Upload the version of your thesis or dissertation approved by your advisor or committee. 
+      You can only upload one file. This file should not contain any signatures or other personally identifying 
+      information because it is a better version for preservation, we recommend you upload a PDF/A file, but it is not required.
```
val99erie: Should there be a period after 'personally identifying information'?  The 'because it is a better version' seems like the start of a new sentence.

little9: :+1:

jenlindner:
@little9 I added a few copy edits

PR #1247, opened 2018-07-04
Add Committee Member selection to the new form

little9:
This committee adds the ability to add
multiple committee chairs and committee members to
the form.


jenlindner:
@little9 we might want to refine the behavior a little but it's important to have the foundation of it in, and I'm going to merge. Thank you.

PR #1249, opened 2018-07-05
validate for missing fields as well as empty fields

jenlindner:
Because two inputs on the About Me form require user interaction in order for their data to be sent to the server, a user can submit the About Me tab without the keys the validation expects. Rather than require anything more special on the front end, this code expects this situation to occur and treats it as an invalid case, the same as if those keys had been sent without data values. I am hoping this is the correct pattern to use for the rest of these validators as well.

### DIFF
--- a/app/validators/about_me_validator.rb
+++ b/app/validators/about_me_validator.rb
@@ -2,8 +2,11 @@ class AboutMeValidator < ActiveModel::Validator
   def validate(record)
     return unless current_tab?(record)
     ::Hyrax::EtdForm.about_me_terms.each do |field|
-      record.errors.add(field, "#{field} is required") if parsed_data(record)[field.to_s].empty?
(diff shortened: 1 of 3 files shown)


app/validators/about_me_validator.rb
```diff
@@ -4,6 +4,9 @@ def validate(record)
     ::Hyrax::EtdForm.about_me_terms.each do |field|
       record.errors.add(field, "#{field} is required") if parsed_data(record)[field.to_s].empty?
```
val99erie: If you change this to `blank?` instead of `empty?`, it will work for the case where the value is `nil`.  Then you can delete that `missing_fields` method and that `rescue` and your test will still pass.

jenlindner:
@val99erie is this good for a merge?

val99erie:
@jenlindner I think you can delete that `rescue NoMethodError` block and the `missing_fields` method.  That code is no longer needed, right?

Issue #1167, opened 2018-06-28
Collection types missing translations


mark-dce:
**STEPS TO REPRODUCE**
1. Login as an administrator
1. Navigate to the dashboard
1. Click on "Settings"
1. Click on "Collection Types"
1. View the "Name" field under "Current Collection Types"

Example: https://qa-upgrade-etd.emory.edu/admin/collection_types?locale=en

val99erie:
A screenshot of the problem:

<img width="810" alt="screen shot 2018-06-28 at 3 50 09 pm" src="https://user-images.githubusercontent.com/298865/42060033-11c29140-7aeb-11e8-80d6-201317b707c6.png">

no-reply:
This seems like it should be coming for free from: https://github.com/samvera/hyrax/blob/1677d72041ecca2e4e88917236dac82f43776b2f/config/locales/hyrax.en.yml#L499

no-reply:
This seems like it should be coming for free from: https://github.com/samvera/hyrax/blob/1677d72041ecca2e4e88917236dac82f43776b2f/config/locales/hyrax.en.yml#L499

val99erie:
@no-reply Yeah, I think you are right.

I haven't been able to reproduce the problem on master yet, and I can't easily confirm whether it was broken in a previous version because the version that is currently deployed on that environment has a SHA from before the force-push on master.  I'm not sure what SHA that corresponds to on the new master.

I'll dig around a little more, but I suspect that I'm not able to reproduce this in my dev environment.

val99erie:
On that environment, here's the last entry from the revisions.log:
`Branch master (at 098e707d7a88507f4fa4dfa8b1ff04d34643588f) deployed as release 20180627180754 by mark`

mark-dce:
Just deployed master and I'm still seeing this text:
```
translation missing: en.hyrax.collection_type.default_title
translation missing: en.hyrax.collection_type.admin_set_title
```

```
04:27 deploy:log_revision
      01 echo "Branch master (at 245e84eade10bb5edc1865107ea16b7fcb2e3652) deployed as release 20180628213126 by mark" >> /opt/laevigata/revis…
    ✔ 01 deploy@qa-upgrade-etd.emory.edu 0.199s
```

mark-dce:
[...]

PR #1250, opened 2018-07-05
Rake task to fix collections with bad (missing translation) labels

bess:
Connected to https://github.com/curationexperts/laevigata/issues/1167

Obviously we would rather fix this upstream in Hyrax, but in the meantime we need a local fix. 

Upstream Hyrax bug at https://github.com/samvera/hyrax/issues/3136

Tested in production mode on `staging-upgrade-etd`. See https://staging-upgrade-etd.emory.edu/dashboard/collections?locale=en

### DIFF
--- a/lib/tasks/fix_collection_type_labels.rake
+++ b/lib/tasks/fix_collection_type_labels.rake
@@ -1,0 +1,15 @@
+# Run this if you are seeing Collections with a type of "translation missing..."
+# Check YOUR_SERVER//dashboard/collections?locale=en for a place where this might appear
+namespace :laevigata do
+  desc "Fix missing collection type labels"
+  task :fix_collection_type_labels do
+    collection_types = Hyrax::CollectionType.all
+    collection_types.each do |c|
+      next unless c.title =~ /^translation missing/
+      oldtitle = c.title
+      c.title = I18n.t(c.title.gsub("translation missing: en.", ''))
+      c.save
+      puts "#{oldtitle} changed to #{c.title}"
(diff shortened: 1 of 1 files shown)


lib/tasks/fix_collection_type_labels.rake
```diff
@@ -0,0 +1,15 @@
...
+# Check YOUR_SERVER//dashboard/collections?locale=en for a place where this might appear
+namespace :laevigata do
+  desc "Fix missing collection type labels"
+  task :fix_collection_type_labels do
+    collection_types = Hyrax::CollectionType.all
+    collection_types.each do |c|
+      next unless c.title =~ /^translation missing/
+      oldtitle = c.title
+      c.title = I18n.t(c.title.gsub("translation missing: en.", ''))
+      c.save
```
val99erie: If you use `c.save` without checking the return value, if there is any problem saving the new title, you won't get any feedback that something is wrong.

For a small number of CollectionTypes where you can easily visually confirm the changes by reloading the page, it probably doesn't matter.

If you want to give feedback if a record failed to save, you could change it to something like:

```ruby
if c.save
  puts "#{oldtitle} changed to #{c.title}"
else
  puts "unable to save new title"
end
```

Or you could use `save!` instead of `save`, but that would abort your loop on the first failure.

I think it's ok to merge this as-is though, because it's not likely to cause problems if one of the titles fails to save.  I'll merge this PR, and you can update the rake task later if you want to add feedback for failed save.

PR #1252, opened 2018-07-05
Cleans up copyrights and patents questions for ETD submission

devanshu-m:
The copyrights and patents questions now have responses as long sentences when previewing an ETD instead of truefalsetrue. Also, the approver will now see Yes or No for those questions instead of true or false.

This PR is supposed to be against 1.6-stable


bess:
@devanshu-m This looks fine to me. Just checking, though... you know this will have to be re-done on the 2.x branch, right?

devanshu-m:
@bess yes, I am aware I will need to rename the references when merging into the 2.x branch. I will make sure everything is more consistent when merging to the 2.x branch.

PR #1263, opened 2018-07-06
Correct the indexing for committee members

little9:
This commit fixes the indexing for names on the
form. It reuses the ID that is generated for them.

Previously, this was using the index from the loop, but this wasn't returning a number in a way I would expect.


little9:
@val99erie this fixes the issue you were seeing with the indexes earlier

PR #1265, opened 2018-07-06
find current in_progress_etd

jenlindner:


app/controllers/in_progress_etds_controller.rb
```diff
@@ -3,8 +3,9 @@ def new
     @in_progress_etd = InProgressEtd.find_or_create_by(user_ppid: current_user.id)
   end
 
+  # TODO: this is effectively always an update. Should we just use the update action instead?
   def create
-    @in_progress_etd = InProgressEtd.new
+    @in_progress_etd = InProgressEtd.find_by(user_ppid: current_user.id)
```
val99erie: Since we are finding by ppid, this should be `current_user.ppid` instead of `current_user.id`.  If I remember correctly, they are always the same.  I'll merge this for now, but we should update it.

Issue #1201, opened 2018-06-28
Clean up deprecations


val99erie:
**Tasks:**

Fix the code so that we don't get these warnings when we run the specs: 

- [ ] ```[DEPRECATION] calling `CurationConcern.actor` with arguments is deprecated and removed from Hyrax 2.0.0. Pass a `Hyrax::Actors::Environment` to `#create`, `#update` or `#delete` instead.```

- [ ] ```[DEPRECATION] caling `Hyrax::Actors::ActorStack` is deprecated and removed from from Hyrax 2.0.0. Use `Hyrax::DefaultMiddlewareStack` instead.```

**Note:**

There is a 3rd warning message when you run the specs, `Passing 'script' command to redis as is`.  We probably cannot get rid of this warning until a future hyrax upgrade.  See discussion here:
https://github.com/samvera/hyrax/issues/1671

PR #1267, opened 2018-07-06
Actor stack refactor

bess:
More syntax updates

Connected to #1201


no-reply:
If these commits all have the same description, do we want to squash them? Or is there more to say about them that we could get into the log?

bess:
I'm happy to squash them.

PR #1268, opened 2018-07-06
Adding --profile to rspec

bess:
If we consistently profile rspec it will help us know
which are our slowest tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,3 +6,7 @@ require_relative 'config/application'
 Rails.application.load_tasks
 
 require 'solr_wrapper/rake_task' unless Rails.env.production?
+
+RSpec::Core::RakeTask.new(:spec) do |t|
+  t.rspec_opts = '--profile'
+end


Rakefile
```diff
@@ -6,3 +6,7 @@ require_relative 'config/application'
 Rails.application.load_tasks
 
 require 'solr_wrapper/rake_task' unless Rails.env.production?
+
+RSpec::Core::RakeTask.new(:spec) do |t|
+  t.rspec_opts = '--profile'
```
val99erie: Can we put a guard on this so it doesn't always run?  I do TDD, and in general I don't want anything extra running during specs (such as coverage or profiles) as it slows down the specs.

If the full suite and individual specs run just as fast with this option, then I don't care, but if it's slower then I need to be able to turn it off.

bess: This will only run if you're running the spec rake task. Is that how you usually run it?

PR #1274, opened 2018-07-09
Refactor to use updated Actor Stack syntax

bess:
Connected to #1201

### DIFF
--- a/spec/features/virus_workflow_etd_spec.rb
+++ b/spec/features/virus_workflow_etd_spec.rb
@@ -25,7 +25,8 @@ RSpec.feature 'Virus checking', :perform_jobs, :clean, :js, integration: true do
 
       attributes_for_actor = { uploaded_files: [upload1.id] }
       env = Hyrax::Actors::Environment.new(etd, ::Ability.new(depositing_user), attributes_for_actor)
-      Hyrax::CurationConcern.actor.create(env)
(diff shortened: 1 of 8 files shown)


spec/features/virus_workflow_etd_spec.rb
```diff
@@ -25,7 +25,8 @@
 
       attributes_for_actor = { uploaded_files: [upload1.id] }
       env = Hyrax::Actors::Environment.new(etd, ::Ability.new(depositing_user), attributes_for_actor)
-      Hyrax::CurationConcern.actor.create(env)
+      middleware = Hyrax::DefaultMiddlewareStack.build_stack.build(Hyrax::Actors::Terminator.new)
```
no-reply: Do these changes remove deprecation warnings?

I'm thinking that the `Hyrax::CurationConcern.actor.create(env)` is a cached instance of `Hyrax::DefaultMiddlewareStack.build_stack.build(Hyrax::Actors::Terminator.new)`, and I feel like I'm missing something about what these changes are doing.

bess: @no-reply Oh, that's totally possible! I was trying to be consistent and copy what I was seeing in the Hyrax rspec suite, but I do not totally understand what I'm doing here. I would welcome a better explanation of what changed in the actor stack from Hyrax 1 -> 2. It maybe be that some of these changes aren't necessary.

PR #1275, opened 2018-07-09
Mock Axios for Jest tests

little9:
This commit adds mocks for Axios so that we
don't get warnings in the jest tests that http
requests haven't been resolved.

This adds some JSON to to the specs that is what
we would expect to get from QA.


bess:
@little9 This isn't passing CI because I think it needs to re-base against master. It's breaking b/c database migrations haven't been run. I added that step to travis, but I think this build was submitted before that? I'm trying to debug it on a couple of other PRs.

little9:
@bess  I rebased and pushed. It looked like one of the builds was using the new method, but another one was using the old method. Now the pr and push builds are using the new method so it should be fine.

Issue #1185, opened 2018-06-28
Switch to the next tab when the previous tab is valid 


little9:
This should be implemented after #1184

PR #1282, opened 2018-07-10
navigating form from left to right, keeping valid tabs enabled

jenlindner:
This commit provides client-server interaction that progresses a user through the form, validating each tab. The form loads with the current tab, and any previous tabs the user has completed are editable (their tabs are clickable), but all tabs that are not yet current are disabled. It is the valid save of the current tab that loads the next tab as current. Addresses #1185

It needs some tests, and there are comments specing out what those should be.

### DIFF
--- a/app/controllers/in_progress_etds_controller.rb
+++ b/app/controllers/in_progress_etds_controller.rb
@@ -4,11 +4,14 @@ class InProgressEtdsController < ApplicationController
   end
 
   # TODO: this is effectively always an update. Should we just use the update action instead?
+  # TODO: this needs to be authorized
   def create
     @in_progress_etd = InProgressEtd.find_by(user_ppid: current_user.id)
-    @in_progress_etd.data = prepare_etd_data.to_json
+    @in_progress_etd.data = prepare_etd_data(@in_progress_etd).to_json
+
     if @in_progress_etd.save
-      redirect_to @in_progress_etd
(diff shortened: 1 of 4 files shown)


app/controllers/in_progress_etds_controller.rb
```diff
@@ -4,11 +4,14 @@ def new
...
   def create
     @in_progress_etd = InProgressEtd.find_by(user_ppid: current_user.id)
-    @in_progress_etd.data = prepare_etd_data.to_json
+    @in_progress_etd.data = prepare_etd_data(@in_progress_etd).to_json
+
     if @in_progress_etd.save
-      redirect_to @in_progress_etd
+      # TODO: we'll want all the json data sent back
+      render json: { lastCompletedStep: current_step, tab_name: tab_name }, status: 200
     else
```
val99erie: If we aren't saving `lastCompletedStep` or `currentStep`, how can you know which tabs should be disabled if a student comes back to edit an existing `InProgressEtd` after a few days?

val99erie: Or would that be handled by re-validating all tabs the first time the form is drawn?

val99erie: After thinking for a few minutes, I don't think it would be a good idea to try to store the current step.  I think it would be best to re-validate all tabs when the form is initialized.   I'll change my review to 'approved'

PR #1287, opened 2018-07-10
Don't hard-code the database config.

val99erie:
The problem I'm trying to solve with this change is unwanted coupling
between environments.  When another developer makes a change to the
config of their local dev environment, my dev environment should be
unaffected by that change.

@bess Could you confirm whether or not this change would require changes to the provisioning and/or deploy scripts?  For example, do we need to put a `config/database.yml` into the shared dir on the environments, or does the deployment already handle the database config?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ coverage
 /node_modules
 yarn-debug.log*
 .yarn-integrity
+
+# Don't hard-code the config
+config/database.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ jdk:
 rvm:
   - 2.4.2
 before_script:
+  - cp config/database.yml.sample config/database.yml
   - bundle exec rake db:create
 script:
   - RAILS_ENV=test ./bin/rails db:migrate
--- a/bin/setup
+++ b/bin/setup
@@ -18,10 +18,10 @@ chdir APP_ROOT do
   system! 'gem install bundler --conservative'
   system('bundle check') || system!('bundle install')
 
-  # puts "\n== Copying sample files =="
-  # unless File.exist?('config/database.yml')
-  #   cp 'config/database.yml.sample', 'config/database.yml'
-  # end
+  puts "\n== Copying sample files =="
+  unless File.exist?('config/database.yml')
+    cp 'config/database.yml.sample', 'config/database.yml'
+  end
 
   puts "\n== Preparing database =="
   system! 'bin/rails db:setup'
--- a/config/database.yml.sample
+++ b/config/database.yml.sample
@@ -1,6 +1,3 @@
-# We run postgres in production so we should run
-# postgres in development and test environments too
-#
 default: &default
   adapter:  postgresql
   encoding: unicode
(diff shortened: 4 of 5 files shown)


bess:
Well, we have been trying to keep as many files as possible in version control, so yes, we wanted to keep `database.yml` in version control too. Is this something we could manage via environment variables? Is there some way `database.yml` itself can be static but environment variables can be set to something local?

val99erie:
Environment variables don't solve my problem.  The problem is that any time someone edits the `config/database.yml`, my environment unexpectedly breaks.  I need to be able to fully config the database, which means that I might need different keys in the config.  Env variables only allow config of values, not keys or declared environments.

Another problem I want to avoid (which is not solved by env variables), is that when another dev changes their env config, that change should be completely isolated.  My environment should be unaffected by config changes in someone else's env.  Checking in the `database.yml` couples my environment to others' environments.

Can we just have the deploy script copy the sample file to the `config/database.yml`?  Then you can keep the production/QA envs as configured by environment variables, but I still have full control over configuration in my own environment.

bess:
By chance, we are actually generating a dot-env enabled database.yml file on the servers already, because that's part of our general ansible build. If we add `database.yml` to the capistrano deploy (here: https://github.com/curationexperts/laevigata/blob/master/config/deploy.rb#L29) then the server will use that copy of database.yml instead, and it won't matter if there isn't one committed here. 

Val, if you add that line to deploy.rb then I'm fine with this change.

val99erie:
Ok sounds good.  I think you are saying that I need to add something like this, right?
`append :linked_files, "config/database.yml"`

I'll add that and re-push

bess:
Yes, exactly.

mark-dce:
Hey - I'd strongly vote that we make the particular configuration we want to vary between environments (either servers or between dev / test / and prod-like) be coded as an environment variable rather than linking files in capistrano.

I.E. instead of this pattern


let's use this pattern https://github.com/curationexperts/laevigata/blob/master/config/solr.yml#L3

mark-dce:
It looks like we already have my proposed solution in place for production, could we just mirror https://github.com/curationexperts/laevigata/blob/master/config/database.yml#L21-L24 for development and test?

val99erie:
@mark-dce That solution doesn't solve my problems.  This has been causing me pain many times, over and over, and I'd like to just fix it for good this time.  I really need full configuration.  Partial config doesn't help me.  I feel frustrated that I have to justify my ability to configure my own environment.

mark-dce:
@val99erie I'm not asking you to justify your need to config your own environment. I'm totally supportive of that.  I'm asking us to find a way that doesn't stomp on my needs from a production support standpoint.

We actually spent significant time getting rid of Capistrano shared files in our deployment configurations because they make diagnosing production problems more difficult and force us to log into the specific server to inspect the actual config - which is a generally bad and unscalable practice. 

Let's figure out a solution that accommodates both our needs.

mark-dce:
@val99erie I really like https://github.com/curationexperts/laevigata/pull/1291/files and I have no problems if we need to add additional keys to that pattern.

PR #1289, opened 2018-07-11
Cleans up copyrights and patents questions for ETD submission

devanshu-m:
The copyrights and patents questions now have responses as long sentences when previewing an ETD instead of truefalsetrue. Also, the approver will now see Yes or No for those questions instead of true or false.

(cherry picked from commit 9c07405549006c73627a7ff70807b2b917edf883)

This PR is supposed to be against master.


devanshu-m:
This PR attempts to find a solution for the issue here: https://github.com/emory-libraries/etd-issues/issues/42

bess:
@devanshu-m I'm going to close this PR for now. Let's see if it's still relevant once the new UI work is done.

PR #1291, opened 2018-07-11
Database config

val99erie:


config/database.yml
```diff
@@ -1,11 +1,13 @@
 # We run postgres in production so we should run
 # postgres in development and test environments too
-#
+
 default: &default
-  adapter:  postgresql
+  adapter:  <%= ENV['DATABASE_ADAPTOR'] || 'postgresql' %>
```
bess: Could we use `DATABASE_ADAPTER` instead? [Either spelling is technically correct](https://writingexplained.org/adapter-vs-adaptor-difference), but I think we should be consistent.

val99erie: yeah, i'll change that

PR #1297, opened 2018-07-12
Fix deploy

bess:
Both of these changes are necessary to fix our capistrano deploy.

1. Wrap entire CI task in a block that excludes it from production (because rspec etc aren't installed in production)
2. Explicitly declare the database adapter for production. Not sure why this is necessary, but it's probably a good idea since we wouldn't want to accidentally override our production database adapater.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,7 +2,6 @@
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
 require_relative 'config/application'
-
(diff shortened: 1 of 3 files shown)


bess:
I'd really like to merge this even though CI is failing. One test, in the js suite, is failing. I can document the failure, but this PR fixes the ability to deploy, which I think is pretty urgent.

Issue #1302, opened 2018-07-12
Disable click events on grayed-out tabs


little9:
Right now they are displayed correctly, but you can still click on them.

PR #1308, opened 2018-07-12
disabling clicks on disabled tabs

jenlindner:
Users should not be able to navigate to disabled tabs, and this commit prevents them from doing so. Addresses #1302


little9:
:+1:

PR #1324, opened 2018-07-16
Add File Upload UI

little9:
This commit adds all the basic functionality
required for the file upload UI. Box features
will be added in a later commit.

This splits out the file upload and delete methods
into their own classes so they can be tested in
an easier way.


jenlindner:
This is fantastic! Super solid code, really readable and structured well, really nice UI.

Issue #430, opened 2017-07-05
Enforce embargo on FileSet view page


bess:
If an embargo is in place, don't allow an unauthorized user to go directly to the show page for the embargoed file set, even if they somehow know the noid of that file set

mark-dce:
@bess @shanitajohn - Isn't this kind of a security risk that we need to prioritize even if Emory hasn't?  I.E. does this become a critical urgent priority the minute files are exposed that should be embargoed?

mark-dce:
Is this equivalent to supporting the default Hyrax embargo cascade functionality and ensuring something similar happens in the embargo job?

PR #1327, opened 2018-07-16
Apply pre-graduation embargoes to added `FileSets`

no-reply:
Adds a `PregradEmbargo` actor that sets a pregraduation embargo in the event that an `:embargo_length` is set. We use it in combination with the default `InterpretVisibilityActor` to get actor behavior that is closer to Hyrax's default behavior.

Specifically, we confirm that embargoes are applied to `FileSet`s when they are attached to a work through the actor stack.

Connected to #430.


no-reply:
Supersedes #1323

no-reply:
@bess the added commit is https://github.com/curationexperts/laevigata/pull/1327/commits/28a56b70ac62f058a6e95a793233ee694bb2b508

no-reply:
The break commented out by 2c7010a is due a latent bug, revealed by running more `save` (and `before_save` hook) calls in the backround jobs inline.

The current code manages to update `committee_members_names` but not `committee_members`. Names are updated to align with the nested committee member nodes on the next save. I think this is a blank node issue.

PR #1346, opened 2018-07-18
Add Box UI to form

little9:
This commit adds the box file picker to the form
and sets up OAuth login. To get this running locally
you'll need to setup the `BOX_OAUTH_CLIENT_ID` and
`BO_OAUTH_CLIENT_SECRET` env variables.

When you click on 'Add You Thesis From Box' you will
be redirected to the Box login and then redirected
back to the submission form. A Box file picker
UI will then be in the UI.

When you pick a file the `box_redirect_controller` will
communicate with the Box servers and get the correct
download URL.

Future commits will use that URL to add the file to the
`InProgressEtd`.

### DIFF
--- a/app/controllers/box_auth_controller.rb
+++ b/app/controllers/box_auth_controller.rb
@@ -1,0 +1,22 @@
+class BoxAuthController < ApplicationController
+  def auth
+    @response = HTTParty.post("https://api.box.com/oauth2/token",
+                              body: {
+                                'grant_type': 'authorization_code',
+                                'code': params[:code],
+                                'client_id': ENV['BOX_OAUTH_CLIENT_ID'],
+                                'client_secret': ENV['BOX_OAUTH_CLIENT_SECRET']
+                              },
+                              headers: {
+                                'Content-Type:' => 'application/x-www-form-urlencoded'
+                              })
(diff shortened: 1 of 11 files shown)


app/controllers/box_auth_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+                              },
+                              headers: {
+                                'Content-Type:' => 'application/x-www-form-urlencoded'
+                              })
+    redirect_to "/in_progress_etds/#{params[:state]}/edit?locale=en&access_token=#{@response.parsed_response['access_token']}"
+  end
+
+  private
+
+    def box_auth_params
```
val99erie: It looks like you aren't using this method anywhere.

app/javascript/Files.vue
```diff
@@ -121,6 +128,36 @@ export default {
...
+      container: '#box-picker',
+      maxSelectable: 1,
+      canUpload: false,
+      logoUrl: 'box',
+      canCreateNewFolder: false
+    })
+  
+    filePicker.addListener('choose', (event) => {
+      console.log(event)
+      var boxRedirect = `/file/box`
```
val99erie: Can you explain what this path is?  `/file/box`.  Is it a placeholder until you get a real URL?

Issue #1347, opened 2018-07-18
File submission from box breaks in old UI


bess:
I'm pretty sure this is only happening for supplemental files submitted from box, and that it's around applying their metadata. 

Linked to Fresh Desk ticket https://curationexperts.freshdesk.com/helpdesk/tickets/65

I thought I had a stack trace for this, but I don't. That stack trace was because I had mistakenly deployed the 2.x branch on qa-etd.

bess:
In the 2.x branch, this bug is caused because in a controller, `params` used to be a sub-class of `Hash`, and now it is an `ActionController::Parameters` object. When you call `.to_h`methods on params now, it checks to see whether the parameters have been whitelisted.

We have custom code that merges two `selected_files` hashes, because the legacy form had two copies of browse everything, so if you submit a PDF and a supplemental file from browse everything, you have to merge those values. 

This fix is low-priority because we hope we will not be running this code in the future. However, if we need to go back and use the non-Vue form, we'll need to fix this before box submission will work.

mark-dce:
@bess can we close this as addressed in the UI refactoring in v2.0.0?

PR #1352, opened 2018-07-19
Try to fix intermittent box upload error

bess:
Connected to #1347

### DIFF
--- a/app/controllers/hyrax/etds_controller.rb
+++ b/app/controllers/hyrax/etds_controller.rb
@@ -150,15 +150,15 @@ module Hyrax
     def merge_selected_files_hashes(params)
       selected_files = {}
       be_files = params.fetch('selected_files', false)
-
+      return unless be_files
       count_of_files = 0
       index = 0
 
       # Get count of all browse-everything selected_files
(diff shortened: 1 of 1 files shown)


app/controllers/hyrax/etds_controller.rb
```diff
@@ -150,15 +150,15 @@ def get_filename_for_uploaded_file(uploaded_file, params)
...
       selected_files = {}
       be_files = params.fetch('selected_files', false)
-
+      return unless be_files
       count_of_files = 0
       index = 0
 
       # Get count of all browse-everything selected_files
-      be_files[0].each_key { |k| count_of_files += be_files[0][k].size }
+      be_files[0]&.each_key { |k| count_of_files += be_files[0][k].size }
```
no-reply: The error I'm seeing in the log is that `be_files[0]` is an `ActionController::Parameters`, which doesn't have an `#each_key. Will this fix that?

If not, maybe we want to try `be_files[0]&.to_h&.each_key`?

bess:
Going to take another swing at this in 2.x. This is a red herring.

Issue #1056, opened 2018-05-31
Ensure UI accessibility requirements are met


mark-dce:

mark-dce:
**ISSUE**
As a user of assistive technologies I would like to be able to use keyboard navigation to submit my ETD.

**Concerns**
- [ ] Hitting the enter key on any tab does a search (unexpectedly) and discards all my progress to-date
- [ ] On the home page I can't tab select the "Submit My ETD" button
- [ ] "Student Name" should have the focus when I start the "About Me" tab
- [ ] "Add Another Committee Chair" should follow the "Committee Chair/Thesis Advisor" in the tab order
- [ ] "I do not have committee members" checkbox should follow "Add Another Committee Chair"
- [ ] "Add Another Committee Member" should be in the tab order
- [ ] There is no obvious way for me to navigate to the "My ETD" tab with a keyboard
- [ ] "Title" should have the focus when I go to the "About my Thesis or Dissertation" tab
- [ ] "Add another Research Field" is not in the tab navigation order
- [ ] "Add another Keyword" is not in the tab navigation order
- [ ] "Copyright" radio buttons are not in the tab navigation order
- [ ] "Permission" radio buttons are not in the tab navigation order
- [ ] "Patents" radio buttons are not in the tab navigation order
- [ ] There is no obvious way for me to navigate to the "My PDF" tab
- [ ] I can't upload a PDF using keyboard navigation
- [ ] There is no obvious way for me to navigate to the "Supplemental Files" tab
- [ ] The "I have no supplemental files" checkbox is not in the tab order
- [ ] The supplemental files checkbox is undersized for users with fine motor challenges
- [ ] I can't upload supplemental files using keyboard navigation
- [ ] There is no obvious way for me to navigate to the "Embargoes" tab
- [ ] The "I do not want to embargo my thesis or dissertation" checkbox is not in the tab order
- [ ] The no embargo checkbox is undersized for users with fine motor challenges
- [ ] There is no obvious way for me to navigate to the "Review & Submit" tab
- [ ] There is no obvious way for me to keyboard select the "Generate Preview" button
- [ ] The "Agreement" checkbox is not in the tab order
- [ ] The "Agreement" checkbox is undersized for users with fine motor challenges
- [ ] The "Submit my ETD" button is not in the tab order

PR #1365, opened 2018-07-20
Ensure the form is keyboard accessible

little9:
This commit changes some controls in the Files
component and adds additional configuration to
Quill to ensure that the form is fully keyboard
accessible. After this commit it is possible to fill
out the entire form using just the keyboard.

Related to #1056

### DIFF
--- a/app/javascript/App.vue
+++ b/app/javascript/App.vue
@@ -39,8 +39,7 @@
             </div>
             <div v-else-if="input.label === 'Table of Contents'">
               <label>{{ input.label }}</label>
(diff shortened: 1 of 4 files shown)


app/javascript/App.vue
```diff
@@ -163,6 +162,12 @@ export default {
       files: [],
       deleteableFiles: [],
       editorOptions: {
+        modules: {
+        toolbar: quillToolbarOptions(),
+        keyboard: {
+          bindings: quillKeyboardBindings()
```
jenlindner: Beautiful

Issue #1206, opened 2018-06-29
A work's visibility should indicate what content is restricted


mark-dce:
**ISSUE**
Currently we only have one type of access restriction in place for all forms of embargo.  But we are really enforcing three different types of embargoes:
* Files only
* Files + TOC
* Files + TOC + Abstract
All of these are also different from an object which should be completely restricted (i.e. not appear in public searches at all)

Any new developer I'll only see one level of access restriction ( and not realize that I have to look at the work itself to determine if there are additional embargo restrictions.  This is OK, but not ideal, in a stand-alone ETD application.  Long-term, however, Emory wishes to use a single repository for all their applications and developers of other applications may not understand that they have to look at metadata within the ETD itself to determine exactly how to enforce the embargo.

Going forward, we should express the different embargo constraints as separate embargo classes, so that a developer unfamiliar with the ETD application would have a clearer indication that TOC and TOC+Abastracts need to be embargoed.

As a model, we can look at the way different visibility levels are handled in [alexandria](https://github.com/curationexperts/alexandria-legacy/blob/master/app/models/admin_policy.rb#L3-L9) vs. the three default levels in [hydra-access-controls](https://github.com/samvera/hydra-head/blob/master/hydra-access-controls/app/models/concerns/hydra/access_controls/visibility.rb#L9-L15).  We should look particularly at the difference between "Restricted" and "Discoverable" in alexandria.

**ACCEPTANCE**
- [x] The application supports three additional visibility levels for
    - [x] files only restricted from public discovery
    - [x] file plus toc restricted from public discovery
    - [x] files, toc, and abstract restricted from public discovery
- [x] keep the existing private visibility level
- [ ] The `README_embargoes.md` file is updated with correct information about how embargoes work (or at least make sure that `README_embargoes.md` doesn't have incorrect info.  Delete it if necessary).

PR #1381, opened 2018-07-20
Set embargo and visibility on FileSet when adding to a work

no-reply:
When attaching a file to a work, Hyrax sets an embargo on the FileSet, matching
the work's permissions. However, we need to determine and set the FileSet
embargo based on the work's status, not simply its visibility. Specifically:

  - we do not want to set FileSet embargo when `work.files_embargoed` is false
  - when we *do* set an embargo, we want to set the FileSet visibility to
    `restricted`/Private, instead of `open`.

### DIFF
--- a/app/actors/hyrax/actors/file_visibility_attributes_actor.rb
+++ b/app/actors/hyrax/actors/file_visibility_attributes_actor.rb
@@ -1,0 +1,33 @@
+module Hyrax
+  module Actors
+    class FileVisibilityAttributesActor < AbstractActor
+      def create(env)
+        next_actor.create(env) && attributes_for_file_sets(env)
+      end
+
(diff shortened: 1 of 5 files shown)


no-reply:
> FWIW - if we refactor embargoes as described in #1206, this PR might be moot because there are no cases when files are not embargoed in the application based on current requirements.

I'm sorry, that's not correct.

This PR ensures that file visibility is set as required by #430. Presently, file permissions are enforced only by work permissions, not file permissions. The effect is that if a file_set and file id are leaked to the user, they can download the file. A brute force attack against this vulnerability is also conceivable.

mark-dce:
Sorry, I think we're violently agreeing again.

Issue #1385, opened 2018-07-23
"Submit my ETD" gives empty page


bess:
As a logged in user with an ETD in progress, when I click the "Submit My ETD" button, I get a blank page. If I then click reload, the page loads as expected.

Maybe similar to the blank page bug Jamie fixed? https://github.com/curationexperts/laevigata/pull/1355

PR #1388, opened 2018-07-23
Fix for blank page when clicking Submit My Etd

little9:
Turbolinks conflicts with the loading of the new form, this
just turns off turbolinks for that link. When you click
it you will go to form without it being loaded by Turbolinks.

This changes this for both the old form and the new form, but won't
have any negative consquences. The page will be loaded using
normal browser nagivation (there won't be a loading bar across
the screen when clicking the link).

For #1385


bess:
:+1: I deployed this PR and yes, this fixes the issue.

PR #1399, opened 2018-07-24
Add submission components

little9:


little9:
@jenlindner I need to push this to a different branch. It's the same as the previous commit, with the rebasing error you pointed out gone.

PR #1407, opened 2018-07-26
Show creator name instead of PPID on review screen

### LINKED ISSUE #1406: Review submissions dashboard should show a name, not a ppid
- [x] Under the `depositor` column, the Review Submissions screen should display a name, not a PPID

bess:
The PPIDs for Emory users are particularly opaque and
Emory has asked that we show the user name where ever possible.
Fixes #1406


bess:
Never mind, I am going to fix this in Hyrax.

PR #1412, opened 2018-07-26
reverting to old way of getting formData for uploader temp bug fix

jenlindner:
hey @little9 I doubt this is what you want to do, but getting the formdata in the old way for the uploader fixes the post bug.


jenlindner:
Ok .. one other thing is we set the formstore data with the tab saves, but the file upload event file info will not be in there yet (unless we wanted to call the formstore set method with the upload event and then read data from there), but doing it this way also send param data and the other way doesn't, and i'm not sure why that would be. Guess it's not really important though.

Issue #1190, opened 2018-06-28
Display the filled-out form content on the submission tab


little9:
The form should display all the content that a user has filled out as a preview before submission.

little9:
@jenlindner Is this ticket no longer valid? Do we just want the final tab in the form to notify the user that they successfully filled out the form, and then give them a link to the actual Etd submission form?

jenlindner:
It's valid. There's a basic version of it working now, the tab displays the form data all in one place. We could close this and open a story for styling the tab and adding the agreement input to it.

little9:
- [ ] Files
- [ ] Embargoes

little9:
This still needs to display the files

PR #1418, opened 2018-07-27
Don't display the abstract/TOC as HTML in submit tab

little9:
This commit adds a class that can be used to strip
the html tags out of the abstract and TOC so that
they can be redisplayed on the final submission tab.

Without this, we'll either need to display the content
as actual HTML which has the potential to break the page,
or show the tags.

Related to #1190


little9:
@jenlindner yep, this is just for display

PR #1420, opened 2018-07-27
Remove Rails edit show page

little9:
We ended up not using this for the final submission tab.

Related to #1190


val99erie:
Also, there's a helper method called `summarize_committee_member`.  If that isn't going to be used by any of Jen's new code, we should delete that too, as part of this cleanup

PR #1423, opened 2018-07-27
fixing typos

jenlindner:


jenlindner:
@little9 this now has the work we just looked at, populating the committee members structures without duplicating, etc.

little9:
@jenlindner Do you think these test failures are real?

jenlindner:
@little9 Yes. I just pushed a branch with a fix, making a pr now.

little9:
@jenlindner I think we can close this because of the other committee member PR

Issue #1198, opened 2018-06-28
Implement uploading  primary file from Box on the front-end


little9:
See if there is a library to enable uploading from Box.

mark-dce:
I don't know if either of these are useful:
https://blog.box.com/blog/using-the-box-api-with-javascript/
https://developer.box.com/docs/sdks --> https://github.com/allenmichael/box-javascript-sdk

It doesn't look like file upload is too complicate - so if we want to contemplate a custom view component for box upload, I'd be willing to talk about whether we have time to complete that (it might be faster than other options.)

PR #1452, opened 2018-08-02
Add a class to perform Box uploads

little9:
This commit introduces a JS class that
can be used to perform uploads from box. This is a
two step process that involves making a request
to box to get the URL of the file and then posting
the result of that to the `/uploads` controller.

Related to #1198

### DIFF
--- a/app/javascript/lib/BoxFileUploader.js
+++ b/app/javascript/lib/BoxFileUploader.js
@@ -2,14 +2,14 @@ export default class BoxFileUploader {
   constructor (options) {
     this.boxAccessToken = options.boxAccessToken
     this.event = options.event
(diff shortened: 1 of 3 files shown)


app/javascript/lib/BoxFileUploader.js
```diff
@@ -0,0 +1,39 @@
+export default class BoxFileUploader {
+  constructor (options) {
+    this.boxAccessToken = options.boxAccessToken
+    this.event = options.event
+    this.crsfToken = options.crsfToken
```
val99erie: Typo:  This should be `csrf` not `crsf`

little9:
@val99erie the typo is fixed now

PR #1462, opened 2018-08-02
Use VisibilityTranslator to determine and set `Etd#visibility`

no-reply:
Adds three new visibility levels to `Etd`. These are usable when an embargo is
set.

First we add delegation to a settable `#visibility_translator` (defaulting to an
instance of `VisibilityTranslator`). This allows the `Etd` to infer its
embargoed `#visibility` level based on the existing `*_embargoed` flags. We then
add logic to `VisibilityTranslator` for setting visibility to the three new
levels _iff_ an embargo is set on the object.

Future work will shift existing clients to using the visibility setters, instead
of setting `*_embargoed` flags manually and add the new visibility levels to the
embargo edit pages. We may also want to consider replacing the `*_embargoed`
flags with data on the `Embargo` ActiveRecord objects.

### DIFF
--- a/README_embargoes.md
+++ b/README_embargoes.md
@@ -15,15 +15,22 @@ In Laevigata the expected embargo behavior is:
   * `toc_embargoed`
   * `abstract_embargoed`
 
+  These correspond to three custom Etd visibility values for use with items under embargo:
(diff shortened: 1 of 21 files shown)


bess:
:+1:

PR #1472, opened 2018-08-03
don't merge!

little9:


app/javascript/formStore.js
```diff
@@ -457,6 +467,9 @@ export var formStore = {
     //these needs to be whatever is current
     formData.append(this.etdPrefix('school'), this.getSelectedSchool())
     formData.append(this.etdPrefix('department'), this.getSelectedDepartment())
+
+    formData.append(this.etdPrefix('files[]'), this.getFiles())
```
little9: Does this need to be `uploaded_files`? That's what we are putting in the form when a file is added.

app/javascript/formStore.js
```diff
@@ -408,6 +408,16 @@ export var formStore = {
 
   /* End of Schools, Departments & Subfields */
 
+  getFiles(){
```
little9: Will we get the files in `savedData` as well after that tab is saved and submitted?

app/javascript/formStore.js
jenlindner: I think no, because that needs to be an array of only ::Hyrax::UploadedFile ids. This might be better named FileInfo though, we need both a name and an id for each file in order to display them properly as well as have the reference we need for deleting. We have that infor in the database, but getting it from there in the middle of the tab saves would be overkill.

app/javascript/formStore.js
jenlindner: Yes.

app/javascript/formStore.js
jenlindner: Although actually, we might need it all for the right display (file size etc) and so maybe we just want to send the sharedState.files array entirely, just in a format the server won't mangle.

Issue #1416, opened 2018-07-27
Call our customized actor stack consistently


bess:
There are places in the code where we're calling the DefaultMiddlewareStack instead of using 
`Hyrax::CurationConcern.actor.create(env)`. Now that we have a customized actor stack, we need to use is consistently or we'll get unexpected behavior.

PR #1486, opened 2018-08-06
Use custom actor stack

bess:
Use our local customized actor stack, not the default one, in feature tests.

Connected to #1416


no-reply:
Are there more places where we want to make fixes for similar test code?

bess:
@no-reply I believe so. I'm fixing them as I'm finding them, and I also have a ticket to remind me to do a universal find and replace.

PR #1491, opened 2018-08-07
Handle `InProgressEtd::NO_EMBARGO` values in the actor stack

### LINKED ISSUE #1485: Submission with no Embargo shows as Embargo
- [x] As a student, when I have selected "No Embargo" during the submission process, I expect to see a green "Open Access" badge when viewing my work. Instead, I see a yellow "Embargo" badge.

no-reply:
The `PregradEmbargo` actor is responsible for setting up attributes for a
six-year pregraduation embargo when an `embargo_length` is passed during `Etd`
creation. This works more or less the same as its predecessor, but relies more
on base Hyrax behavior. The values passed here are eventually interpreted at
graduation time to determine the post-grad embargo release date.

However, the `InProgressEtd` data stores a specific string indicating that no
embargo is requested and passes that value to the stack. This is always
interpreted as a request for a pregrad embargo.

Ideally, we would handle this at the edge of `InProgressEtd` so other parts of
the application don't need to be aware of this blessed string. It's likely that
`InProgressEtd` needs this to be round-tripabble, and the handling for
transformation of this data lives across various methods. Rather than attempt an
immediate refactor, we fix the bug by handling the special data in the actor
stack. We raise a warning so the larger fix won't get ignored.

Fixes #1485.

### DIFF
--- a/app/actors/hyrax/actors/pregrad_embargo.rb
+++ b/app/actors/hyrax/actors/pregrad_embargo.rb
@@ -14,6 +14,9 @@ module Hyrax
         def pregraduation_embargo_attributes(env)
           return {} unless env.attributes.key?(:embargo_length)
 
+          return handle_malformed_data(env) if
+            env.attributes.fetch(:embargo_length, nil) == InProgressEtd::NO_EMBARGO
+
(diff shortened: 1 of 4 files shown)


no-reply:
#1492 is related, but better taken on when `InProgressEtd` and the form are more stable.

Issue #1433, opened 2018-07-30
Allow users to see their ETDs pre-graduation


bess:
- [ ] After I've submitted my work, and before I've graduated, I should be able to ~search for my ETD and find it~ see my work under My Works
- [x] There should be a menu item in the top right corner taking me to My ETD (the My Works dashboard)

bess:
My Works has been fixed in Hyrax master. We need to either incorporate that fix or (preferably) get it merged into Hyrax and get a Hyrax point release out

bess:
Waiting on release of Hyrax 2.1.1

PR #1533, opened 2018-08-08
Upgrade to hyrax master

bess:
We are depending on several bug
fixes that have happened in Hyrax
master. Until version 2.1.1 is
released, we need to run against
master so we can get these fixes.

Connected to #1433


bess:
This is stale. Re-opening another.

Issue #1505, opened 2018-08-07
Allow Approvers to manage embargos for their school post graduation


bess:
- [ ] As an approver (not an admin) I can edit the embargo after the student has graduated
- [ ] How do we demo this?

bess:
This definitely isn't working yet. I get an error if I try to go to the embargo page as `candleradmin`

PR #1541, opened 2018-08-09
Fix embargo edit screen

bess:
Do not fail if embargo release date
is nil or if visibility is nil.

Connected to #1505


bess:
embargo_screen_working

PR #1545, opened 2018-08-09
Upgrade3 to hyrax master

bess:


bess:
Rebasing this against master and latest Hyrax master

PR #1575, opened 2018-08-10
Make "hidden" Etds have "private" visibility

### LINKED ISSUE #1561: A work that is hidden should have a "Private" visibility flag
- [ ] It should be clear from looking at the show page of a hidden work that it is hidden. It should display a red "Private" badge.

no-reply:
When an Etd is marked "hidden", we want its visibility to reflect the original implementation throughout the application. This ensures "private" status is represented even when the object is under embargo.

Fixes #1561.

### DIFF
--- a/app/indexers/etd_indexer.rb
+++ b/app/indexers/etd_indexer.rb
@@ -16,7 +16,7 @@ class EtdIndexer < Hyrax::WorkIndexer
 
   class IndexingService < Hyrax::BasicMetadataIndexer
     self.stored_fields +=
-      [:abstract, :email, :post_graduation_email]
+      [:abstract, :email, :post_graduation_email, :hidden?]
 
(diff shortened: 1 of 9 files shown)


no-reply:
@bess I have some slight concerns about the way that "hidden" is currently implemented; note the comment above "regardless of read groups, embargoes, and other visibility/permissions statuses."

Does this accurately reflect the existing implementation? Could we talk about inverting this so `hidden = true` _sets private visibility_, and `Etd#hidden?` is an alias for `#visibility == "restricted"`?

no-reply:
We resolved the comment above over the phone. The Workflow handles setting the correct visibility when hiding the object. We simply pass through to the "proxy" original implementation when `#hidden?` is true to keep the badge correct.

PR #1591, opened 2018-08-11
Show school, department, and embargo length on edit screen

little9:
This commit includes fixes for the edit screen. Mostly
this consists of changing the values that we get back
from Fedora to match what we are expecting.

Fix for language not displaying on the edit form

Display research fields on the edit form

### DIFF
--- a/app/javascript/Language.vue
+++ b/app/javascript/Language.vue
@@ -33,10 +33,6 @@ export default {
     getSelected(data){
       var selected = this.sharedState.getSavedLanguage()
       if (selected !== undefined) {
-        //TODO: the model should return strings not arrays 
-        if(_.has(this.sharedState.savedData, 'etd_id')){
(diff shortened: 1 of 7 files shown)


little9:
After this PR, these still need to be fixed on the edit form:

* There is a problem with the chairs, where they append copies of  themselves in the controller. So they are doubling when you save. 

* Copyright questions. What we get back from Fedora are strings: 'true' or 'false', these need to be 1 or 0

* Files
  This is what Val was working on. When you edit a work the submitted files don't show up. If they edit a  work are we going to require them to upload their files again?

* Embargoes
  What we get back from fedora is not what is in the form. So we either need to translate between the to
  or change the form so that it matches fedora expects.

bess:
@little9 Would you please translate the above comments into tickets?

Issue #1594, opened 2018-08-11
Copyright question response in edit screen different than from InProgressEtd


little9:
We are using 0 and 1 to keep track of the copyright questions. When we get a response from the 
server in the edit screen these have been turned into strings: 'true' and 'false', not boolean values or 0 or 1. We need to switch these around when loading the values into the form on the edit screen.

PR #1597, opened 2018-08-12
Editable copyright questions

little9:
This commits adds the ability to edit
the copyright questions on the edit screen.

Connected to  #1594 
Connected to #1602 
Connected to #1599


bess:
Hi, @little9. For the editable copyright questions, would you please make them display the currently selected choice as the first option? Otherwise, when an admin is editing it, they'll have to remember what was there before and re-select it.

show_selected_options_on_edit_screen

little9:
@bess this is present in the PR, along with the fix for #1594, #1602, and for #1599 they now display true or false instead of 0 or 1.

Issue #1586, opened 2018-08-11
Updating embargo deletes abstract and toc


bess:
- [x] When I update the embargo I expect to see all fields except the embargo as they were before, especially abstract and toc

Issue #1489, opened 2018-08-06
Updating embargo goes to confusing redirect page


bess:
- [x] As a user who can update, embargoes, when I update an embargo from the Manage Embargoes dashboard, I am taken to a confusing screen that appears to be a malformed redirect (see attached image)

bess:
![redirect_error.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5515e9ab52d65b2000638516/125516c66e82c728ace21e0d46fa919e2a9086aa9bd6ba51b652e2d665043761260ea62864c7b96aea1a7f02575949f9460e0b17a8fe892ba7e768778c0803bbc3714d36976345f3e5db55ea43a82a994e8dbcbd8f4fc3ed3524e68b8a4a3d1b1755.png)

PR #1598, opened 2018-08-12
Merge UI update behavior with embargo updates

bess:
Do not re-set abstract and table_of_contents
params unless they exist and need to be sanitized.
Do not short-circut after_update_response by
sending the response prematurely.
Redirect the user to the object show page after
an update, regardless of whether they have updated
via the UI form or the embargo edit form.
Feature test for editing embargoes.

Connected to #1489 
Connected to #1586


bess:
Closing and re-submitting in combined PR.

Issue #1623, opened 2018-08-13
Dept field doesn't display on Submit tab (different than #1622)


RachelDCE:
When I finish and save Embargo - log out, open in new incognito window and log back in - the Department doesn't show up on the summary.
 
I have to click on My Program and back to Submit for the Dept to show in the summary.

Similar to #1622 - However School and Dept don't go missing in the About Me and My Program tabs

![screen shot 2018-08-13 at 11 16 30 am](https://user-images.githubusercontent.com/23509160/44044867-82d28e7e-9eec-11e8-8317-452411f1e7bf.png)

![screen shot 2018-08-13 at 11 16 55 am](https://user-images.githubusercontent.com/23509160/44044881-8b22ba54-9eec-11e8-957e-a080ec214161.png)

![screen shot 2018-08-13 at 11 17 16 am](https://user-images.githubusercontent.com/23509160/44044893-91549082-9eec-11e8-9a76-26b055f85757.png)

RachelDCE:
Tested on qa.upgrade-etd / © 2017 Emory University. Version js/fix-school-dept-embargo-switching updated 14 August 2018.

Issue #1622, opened 2018-08-13
School and Dept disappear at Submit page


RachelDCE:
I entered and saved each page (logging out in the middle of the submission to test in-progress fields were retained) - Everything was retained until the end.

It appears the School and Dept only disappeared once I got to the Submit page. 

I went back and added School and saved. My Dept automatically re-populated.

I went back to Submit page and as it was still working - I had to Refresh in order to submit. But, it did submit at that point.

I will test it again and see if I can recreate it.

RachelDCE:
![screen shot 2018-08-13 at 8 47 50 am](https://user-images.githubusercontent.com/23509160/44037302-9adfae7e-9ed9-11e8-952c-1817f9a0affb.png)

RachelDCE:
Tested on qa.upgrade-etd / © 2017 Emory University. Version js/fix-school-dept-embargo-switching updated 14 August 2018.

Issue #1617, opened 2018-08-12
Can't submit more than one file from box


bess:
If I submit a file from box, then try to submit a second file from box, the first file disappears.

Issue #1603, opened 2018-08-12
Adding a file from box removes the school and department


bess:
During my submission process, if I upload a file from box it removes my school and department.

bess:
@little9 I just tested this with your Sunday night PR and I can re-create it on the server.
1. Start a new ETD
2. On the file submission page add a pdf and a new supplemental file 
3. Go to the next tab and you can't select embargo b/c the school setting has been lost

little9:
@bess Did you clear out any InProgressEtds before testing it?

PR #1624, opened 2018-08-13
Better management of school, dept, and embargo state

little9:
This commit fixes some bugs in the management of the
state of the properties that are based on the school.

The state is based on the saved school primarily now.

A copy of the school is also saved in the browser's
`localStorage` as a precaution.

This commit also adds a  destroy action to to the
InProgressEtd controller, this allows for a Start
Over button to be in the form that can be used
by a user to delete their InProgressEtd and start
over from scratch in case something goes wrong or
they enter the wrong school.

Connected to #1623 
Connected to #1622 
Connected to #1603 
Connected to #1617

### DIFF
--- a/app/controllers/in_progress_etds_controller.rb
+++ b/app/controllers/in_progress_etds_controller.rb
@@ -35,6 +35,14 @@ class InProgressEtdsController < ApplicationController
     end
   end
 
+  def destroy
+    @in_progress_etd = InProgressEtd.find(params[:id])
+    authorize! :update, @in_progress_etd
+    @in_progress_etd.destroy
+
+    redirect_to root_url
+  end
+
   private
 
     def current_step(data)
--- a/app/javascript/App.vue
+++ b/app/javascript/App.vue
@@ -9,7 +9,7 @@
         </button>
       </li>
     </ul>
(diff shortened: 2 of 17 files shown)


bess:
If I do not attempt to submit from box, this works fine.

little9:
@bess I found what was changing the school and departments during the box interaction, so that is fixed in this PR now.

little9:
@bess I pushed a fix for the department. I've got a selenium test for it, so I can show that it doesn't work on the previous commit, but does now

bess:
I'm happy for this to be merged as soon as @RachelDCE can confirm it's working as expected.

little9:
@bess Does the scope for #1617 include adding a file with box, refreshing and then adding another file from box? That case won't work, what will work in this branch is adding a file from Box and then adding another one and then saving. 

I think the first case is covered by #1618, but I made another ticket for particular case: https://github.com/curationexperts/laevigata/issues/1627

bess:
@little9 No, I wouldn't expect to be able to add a file, refresh, and then add another file. I just want to be able to add more than one file from box while I'm on that tab, before I've hit "save an continue"

little9:
@bess Just tried this with Rachel and Dept and School are retain, but this doesn't address #1617 so we shouldn't close that

Issue #1626, opened 2018-08-14
Do not let me leave the file upload screen without a primary PDF


bess:
Currently, I can attach a supplemental file from box before I upload a primary PDF, then click "save an continue" and go to the next screen as if I'm done with my file uploads. It would be really bad if the student didn't upload their primary PDF. Please ensure we're checking for this, preferably before they leave the file upload tab.

PR #1628, opened 2018-08-14
backend primary file validation

jenlindner:
This commit adds model validation ensuring a student can't save the files tab with a primary PDF. It is connected to #1626.

Connected to #1626


jenlindner:
okay @little9 I think it's ready.

Issue #1447, opened 2018-07-31
ETD submission: error case


val99erie:
- [ ] When the student fills in the entire form and submits the ETD for approval, if the `Etd` record fails to save for any reason, the errors should be displayed on the page so that the student can correct the problem.

Background info:  This is stubbed out in the controller here:
https://github.com/curationexperts/laevigata/blob/master/app/controllers/hyrax/etds_controller.rb#L60-L61

OR

- [ ] It needs to at least tell the user that there has been an error. Raise an error if there is an error. It can't just spin forever.

bess:
Testing: Try getting to review screen, then removing all keywords and trying to submit.

val99erie:
I can confirm that with current code on master, if update fails in EtdsController, the form just hangs forever.  No error message, no way to recover.  Current master is `89f1071`

val99erie:
@RachelDCE

I thought of a way you can test this story using only the UI (not having to use the Rails console or change any code):

* When you are viewing an ETD that you have permission to edit, click the 'Edit' button.

* When the form loads, make a change to one of the fields in the form, but don't hit the submit button yet.

* In another browser tab, delete the ETD (you can do it from the show page for the ETD)

* Back in the tab for the edit form, click the 'submission agreement' box, and then hit the button to submit the ETD.

Acceptable behavior:  You see some kind of error message.

Bad Behavior:  The spinner just keeps spinning forever, and you aren't sure if it's working or not.

Correct behavior:  Display a really nice error message that explains exactly what is wrong.  (This is not implemented yet, so we don't expect this behavior at this time)

PR #1637, opened 2018-08-14
Notify user if errors happen during submission

little9:
This commit will catch errors during
the submission and set an error state
in the formStore. That changes
the display from the spinner to an
error notification.

Connected to #1447


little9:
@val99erie this is what we discussed, it displays an error in the form if there is a JS error 
or if the ajax request errors out.

Issue #1675, opened 2018-08-17
When editing Committee Chair & Member - edited chair/member persists and new is added


RachelDCE:
When editing Committee Chair & Member:

I edit the Name field and instead of replacing/changing that name- a new chair/member is added, so that I end up with multiple unintentional chairs/members. New ones added each time I try to edit.

![screen shot 2018-08-17 at 3 33 33 pm](https://user-images.githubusercontent.com/23509160/44287576-ee026900-a232-11e8-9329-ba01b6e7c996.png)

Tested in QA | 2017 Emory University. Version master updated 17 August 2018

val99erie:
The solution that I implemented works properly from the point-of-view of a student, but it leaves behind orphaned records in fedora.  After discussion with the team, since we are so close to the deadline, we decided that this is acceptable for now.  I wrote story #1733 to fix this problem in a future release.

RachelDCE:
Was able to change and add committee chairs and members without unintended multiples showing. 

Tested on QA | 2017 Emory University. Version v2.0.0-beta8 updated 22 August 2018

Issue #1616, opened 2018-08-12
Be able to remove a committee member


bess:
- [x] When editing an ETD be able to remove a committee member or chair

val99erie:
The solution that I implemented works properly from the point-of-view of a student, but it leaves behind orphaned records in fedora.  After discussion with the team, since we are so close to the deadline, we decided that this is acceptable for now.  I wrote story #1733 to fix this problem in a future release.

RachelDCE:
Was able to edit both committee member and chair. 

Tested on QA | 2017 Emory University. Version v2.0.0-beta8 updated 22 August 2018

PR #1679, opened 2018-08-18
Initial work on members/chairs for editing

little9:
The committee chairs/members aren't updating
correctly. When updating an Etd this commit
should remove any existing members/chairs
and replace them with whatever is in your update
submission.

Before members/chairs where never being deleted
but appended to on each update so you would have
many duplicates. Whenever I run this code it seems
like sometimes the update hasn't been indexed
correctly because the members don't show up on
show page, but are present on the `Etd` object.

Connected to #1675
Connected to #1616

### DIFF
--- a/app/controllers/hyrax/etds_controller.rb
+++ b/app/controllers/hyrax/etds_controller.rb
@@ -77,6 +77,8 @@ module Hyrax
       end
 
       if params['request_from_form'] == 'true'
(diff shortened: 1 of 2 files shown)


val99erie:
@little9 I'm closing this PR without merging.  This should be handled by this PR:
https://github.com/curationexperts/laevigata/pull/1732

Issue #1691, opened 2018-08-19
Provide a message if Javascript isn't available


mark-dce:
**ISSUE**
When the user logs in to the application without javascript enabled and click submit, they get a blank screen.

**ACCEPTANCE**
- [x] The default submission screen renders a notice that javascript is required to submit:  "NOTE: The Thesis and Dissertation submission system requires javascript.  Please enable javascript or make your submission using a public computer provided by the University."

bess:
![no_javascript.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5515e9ab52d65b2000638516/125516c66e82c728ace21e0d46e69ba5298395a89ceaad4ab449b2d86b56357c2301a03f32dfa17de746785959581bab170c0c10fbf9892af7e23b76de545dffcd7f1d2598694ef1e1c84afd13b0328a468fb7b98948c7ee3521ef898d4d3f0c48.png)

mark-dce:
👍 

Confirmed on https://qa-upgrade-etd.emory.edu/
Version v2.0.0-beta5 updated 20 August 2018.

PR #1698, opened 2018-08-19
Add a no script tag in the form view

little9:
Adds a noscript tag to inform users that
JS is required. Wording taken from #1691.

Connected to #1691


little9:
noscript

Issue #1696, opened 2018-08-19
Students should be able to enter sub-deparments


mark-dce:
**ISSUE**
On the "My Program" tab I can not enter a subfield for the departments listed in the school reports for [Emory College](https://qa-upgrade-etd.emory.edu/schools/Emory%20College?locale=en),  [Laney Graduate School](https://qa-upgrade-etd.emory.edu/schools/Laney%20Graduate%20School?locale=en), and [Rollins School of Public Health](https://qa-upgrade-etd.emory.edu/schools/Rollins%20School%20of%20Public%20Health?locale=en)

<img width="1189" alt="1696 - dept sub-dept current" src="https://user-images.githubusercontent.com/3064318/44313739-2d67bb80-a3d3-11e8-8ea4-dcd8f07ab92c.png">

**ACCEPTANCE**
- [x] I can enter both department and sub-department for the sub departments listed in the schools reports.

**OBSERVATION**
Since Departments and Sub-departments are tightly coupled and hierarchical, consider rendering them together as [select optgroups](https://www.w3schools.com/tags/tag_optgroup.asp) rather than two separate fields. (the sample code to generate this screenshot is included below)

<img width="1176" alt="1696 - dept sub-dept using optgroups" src="https://user-images.githubusercontent.com/3064318/44313745-43757c00-a3d3-11e8-80b6-0077c0bc703e.png">

```ruby
<select name="etd[department]" id="department" aria-required="true" class="form-control">
    <option value="Behavioral Sciences &amp; Health Education - MPH | ">
      Behavioral Sciences &amp; Health Education - MPH
    </option>
    <optgroup label="Biostatistics and Bioinformatics">
        <option value="Biostatistics and Bioinformatics | Biostatistics - MPH & MSPH">
          Biostatistics - MPH & MSPH
        </option>
        <option value="Biostatistics and Bioinformatics | Public Health Informatics - MSPH">
          Public Health Informatics - MSPH
        </option>        
    </optgroup>
    <option value="Environmental Health | ">
      Environmental Health
    </option>
    <optgroup label="Epidemiology">
        <option value="Epidemiology | Epidemiology - MPH & MSPH">
          Epidemiology - MPH & MSPH
        </option>
        <option value="Epidemiology | Global Epidemiology - MPH & MSPH">
          Global Epidemiology - MPH & MSPH
        </option>
    </optgroup>
    <optgroup label="Executive Masters of Public Health - MPH">
[...]

PR #1704, opened 2018-08-20
Update subfield JSON file to include all options from /schools

little9:
This updates the file that used by the UI to detemine which department options
are associated with a subfield/program vocab.

Related to #1696


mark-dce:
Are we sure this works? it looks like #1700 broke Rollins submissions, so I reopened #1696.

Issue #1701, opened 2018-08-20
Fix spacing of button to remove partnering agency


mark-dce:
**ISSUE**
The visual spacing of the removal button is exactly the same distance between the entry it applies to and the next entry.  This makes it difficult to quickly visually scan and quickly identify which field the button is attached to.

**ACCEPTANCE**
- [x] Place the removal button to the right of the associated field OR at least provide more whitespace between the the button and the non-related item.

Current
===========
<img width="592" alt="1701 - curent spacing" src="https://user-images.githubusercontent.com/3064318/44316182-e89e4d80-a3ef-11e8-8e03-13900df2f0af.png">

Better
===========
<img width="617" alt="1701 - better spacing" src="https://user-images.githubusercontent.com/3064318/44316243-331fca00-a3f0-11e8-9080-43c460f41551.png">

Best
========
<img width="702" alt="1701 - best spacing" src="https://user-images.githubusercontent.com/3064318/44316971-45036c00-a3f4-11e8-9833-94bcd2f3a0a1.png">

bess:
![Screenshot from 2018-08-20 08-30-01.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5515e9ab52d65b2000638516/125516c66e82c728ace21e0d46db978826878dba87e6ab03a24fff99691477227149f3733dd7b368ef1a36525f424ead53444d04bea6d37ef5f22e7a985903b982655973933a47f8ebd618bf01bb71df4883e2b4dc5edcfa7f74b9dfd0126754505a9bf54f34a9934acd5bf24ce74450a7113067e4647b.png)

RachelDCE:
I am still seeing this as before only the buttons appear to be slightly indented.

![Screen Shot 2018-08-20 at 4.40.58 PM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5835b9b16342041300e166e0/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b772b6456f36b7191a37cf1162b4f5a575ecc3c4a4356bcbbd671f3e57862804e0ee5853e5772c16816faecd14bb801ba21da1b82b0e8821ad2f42f67b6d5db1063474f4dcbed5727a19e41c95df54fe94d54a0143366e3702431.png)

little9:
@RachelDCE are you seeing this in all browsers you try?

RachelDCE:
[...]

PR #1705, opened 2018-08-20
Add styling for remove partnering agency buttons

little9:
This moves the remove partnering agency button
to the right of each of these select boxes.

Connected to #1701


bess:
screenshot from 2018-08-20 08-30-01

Issue #1721, opened 2018-08-21
Can't Save and Continue My Files in Safari


RachelDCE:
I've tested in Safari and Chrome (Incognito windows with a new ETD) and I can't save and continue from My Files

Testing on QA / 2017 Emory University. Version v2.0.0-beta5 updated 20 August 2018

Screen recording of issue: https://youtu.be/uhF5sYDYxKc

RachelDCE:
Tested more and was able to get My Files to save in Chrome - Still having the issue in Safari.

RachelDCE:
Tested and fixed

Tested on QA | 2017 Emory University. Version v2.0.0-beta8 updated 22 August 2018.

PR #1722, opened 2018-08-21
Ensure that we don't submit blank file values in SaveAndSubmit

little9:
Connected to #1721


little9:
@jenlindner I changed this like you suggested and the file validation still works and you can continue in Safari.

Issue #1724, opened 2018-08-21
Incorrect Rollins department names


bess:
Submissions are not working for some Rollins departments because they names in the drop down menu don't match the names of the Admin Sets. 

E.g., The dropdown menu says "Behavioral Sciences & Health Education - MPH" but the AdminSet is called ""Behavioral Sciences and Health Education"

- [x] All options in the department drop down menu for Rollins must correspond to the name of an AdminSet as defined in `config/emory/admin_sets.yml`

bess:
@RachelDCE When you're ready to test this, would you please ensure that you can successfully submit for every department that's possible to select for Rollins? I've just done that but I'd appreciate a second pair of eyes on it.

RachelDCE:
Department/subfield names aren't the same between My Program tab and Submit tab

My Program Tab

![Screen Shot 2018-08-23 at 10.54.45 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5835b9b16342041300e166e0/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b772b6456f06b7191a379ef0c2e55415b4bbc30254d58eeb9cb74fce36f34985619e8d9390c7cc03a44abeed64ceb06ba208a1ed1b1bade4496fa2137a5dad11b61435c52dcbd4f3fb29842c259f348e44d56ad1b366ae0743f6e0a.png)

Submit Tab
![Screen Shot 2018-08-23 at 10.55.05 AM.png](https://waffleio-direct-uploads-production.s3.amazonaws.com/uploads/5835b9b16342041300e166e0/125516c66e82c728ace21e0d46db978826878de9bce1b057e40fa0c5710b772b6456f06b7191a379ef0c2e54415f4bbc30254d58eeb9cb74fce36f34985619e8d9390c7cc03a44abeed64ceb06ba208a1ed1b1bade4496fa2137a5dad11b61435c52dcbd4f3fb29842c259f348e44d56ad1b3765e17e366e0a.png)

RachelDCE:
Department/subfield names aren't the same between My Program tab and Submit tab:
My Prog Tab:Behavioral Sciences and Health Education

RachelDCE:
I am able to submit to all Rollins departments. See #1748 for issue with matching department names between MyProgram and Submit tabs. 

Tested on QA | © 2017 Emory University. Version master updated 23 August 2018.

PR #1726, opened 2018-08-21
Change the program label to match admin_set.yml

little9:
Connected to #1724


bess:
admin_set_error

little9:
@bess I just pushed up a change for this. The problem was not using the ID for the value in the department field.

little9:
annotation

When you select `Behavioral Sciences ... MPH` in the form it will be saved with the department that corresponds to to the admin set not one with MPH in it after this PR.

Issue #1677, opened 2018-08-17
Replacing the primary PDF file only works the first time


val99erie:
**Steps to reproduce the problem:**

* Log in as a student, fill in the form for a new ETD, and submit it.

* Log in as an approver, and mark the ETD as 'changes requested'.

* Wait for background jobs to finish running.

* Log in as the student, and press the button to edit the ETD.

* In the edit form, remove the primary PDF file, add a new primary PDF file, and submit the form.

* Wait for background jobs to finish running.

* Reload the ETD show page, and you should see your new PDF file (as expected).

* Hit the edit button again.

* Do it again:  In the edit form, remove the primary PDF file, add a new primary PDF file, and submit the form.

* Wait for background jobs to finish running.

**Notice these problems:**

* You won't see it in the list of files on the ETD show page, but if you click 'File Manager', you'll see the old PDF file that you supposedly deleted.

* The new file did get saved, but the thumbnail and representative are still set to the old file.

* If you look in the Rails log, you'll see 403 errors from CANCAN access denied.  This was from the request to delete the file.

```
[CANCAN] Checking edit permissions for user: valerie with groups: ["public", "registered"]
[CANCAN] edit_groups: []
[CANCAN] edit_users: []
[CANCAN] decision: false
Completed 403 Forbidden
```

* If you check the rails console, you can see that there's one more FileSet record than you would expect, and there's an extra primary PDF file attached to the Etd record:

```
etd.primary_file_fs.count
 => 2   # This should be 1
```

val99erie:
I suspect this same problem will exist for supplemental files - you'll be able to remove files that you added the first time you submit, but if you add files later, probably those files will have this same 'access denied' problem.

bess:
As of version v2.0.0-beta4 updated 20 August 2018, I cannot re-create this error. On the qa-upgrade server I took an existing ETD. I replaced the primary PDF several times in a row. I attached supplemental files and replaced each of them several times in a row. I believe this got fixed by some other change we made. I see the correct behavior in the UI, and when I check the object on the console I also see only the expected file sets attached:
```
irb(main):002:0> etd.file_sets
[...]

PR #1730, opened 2018-08-21
Allow users to replace their files more than once

bess:
The first time a work is returned to the user via the workflow
("request changes"), the edit permissions on all attached files
are set correctly and the files can be replaced. Unfortunately,
that process of setting the file set permissions was not being
called on subsequent edits and that meant the user could only
replace their files once, and thereafter it would fail.

This PR fixes this bug. We belive this is a but in underlying
Hyrax and I'm also going to follow up with a bug report there.

Connected to #1677

### DIFF
--- a/app/jobs/attach_files_to_work_job.rb
+++ b/app/jobs/attach_files_to_work_job.rb
@@ -26,6 +26,15 @@ class AttachFilesToWorkJob < Hyrax::ApplicationJob
       actor.attach_to_work(work, metadata)
       uploaded_file.update(file_set_uri: actor.file_set.uri)
     end
+    # If the Depositor already has edit rights, make sure those get applied to
+    # any newly added files. These methods get called correctly when a work is
+    # first returned to a depositor for changes, but they do not otherwise get called for
+    # subsequent file changes. This fixes a bug where a user can only replace their
+    # files once.
+    if in_edit_state?(work)
+      Hyrax::Workflow::GrantEditToDepositor.call(target: work.reload)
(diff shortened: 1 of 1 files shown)


app/jobs/attach_files_to_work_job.rb
```diff
@@ -26,6 +26,15 @@ def perform(work, uploaded_files, **work_attributes)
       actor.attach_to_work(work, metadata)
       uploaded_file.update(file_set_uri: actor.file_set.uri)
     end
+    # If the Depositor already has edit rights, make sure those get applied to
+    # any newly added files. These methods get called correctly when a work is
+    # first returned to a depositor for changes, but they do not otherwise get called for
+    # subsequent file changes. This fixes a bug where a user can only replace their
+    # files once.
+    if Hyrax::Workflow::PermissionQuery.scope_permitted_workflow_actions_available_for_current_state(user: User.find_by_user_key(work.depositor), entity: work.to_sipity_entity).pluck(:name).include?("request_review")
+      Hyrax::Workflow::GrantEditToDepositor.call(target: work.reload)
```
val99erie: I understand why you are calling the `InheritPermissionsJob` here, but why is it necessary to call `GrantEditToDepositor`?  Shouldn't the parent work's edit permissions already be correct at this point?

bess: @val99erie Yes, the parent work's edit permissions are in place, but the edit permissions on the attached file sets are not. I've tested this a bunch of times and it does not work unless we run that `GrantEditToDepositor` method.

PR #1732, opened 2018-08-21
Edit or remove committee members. Connected to stories #1616 and #1675.

val99erie:
I cleared the committee members and chairs before passing the new params
from the form to the actor.

I fixed the methods that keep the committee_chair_name and
committee_members_names in sync with the committee_chair and
committee_members fields.

### DIFF
--- a/app/controllers/hyrax/etds_controller.rb
+++ b/app/controllers/hyrax/etds_controller.rb
@@ -77,6 +77,8 @@ module Hyrax
       end
 
       if params['request_from_form'] == 'true'
+        curation_concern.committee_chair = nil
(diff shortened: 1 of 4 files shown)


val99erie:
@little9 No, it doesn't matter if you set them to nil or empty array.

The reason it wasn't working for you is that there was a second problem.  For some reason we have these properties: `committee_chair_name` and `committee_members_names`.  (I don't know why these are persisted in fedora instead of just in the solr index).  Those fields need to be kept in sync with the `commitee_chair` and `committee_members` fields, but the methods that updated those fields had bugs in them that prevented the fields from being updated if you delete all the members or chairs.

Issue #1739, opened 2018-08-22
Language not present in Migrated ETDs


mark-dce:
**ISSUE**
When I view a migrated ETD like 
https://staging-upgrade-etd.emory.edu/concern/etds/zc77sq630?locale=en ( see [legacy version](
https://legacy-etd.library.emory.edu/view/record/pid/emory:cqt78) ) the language field is not displayed.  When I edit the ETD, the language field asks me to select a language.

Server: https://staging-upgrade-etd.emory.edu
Version v2.0.0-beta7 updated 21 August 2018

**STEPS TO REPRODUCE**
1. Login to staging-upgrade as an administrator
1. Find any migrated ETD - e.g. zc77sq630
1. Scroll down to the "About this Dissertation" - look for language
1. Click on edit
1. Scroll down to the "About this Dissertation" section

**EXPECTED BEHAVIOR**
* I should see the language displayed on the show view
* The particular language of the submission should be selected

**ACTUAL BEHAVIOR**
* No language is displayed in the show view
* In the Edit view, the Language field has "Select a Language" displayed

SHOW VIEW (Created in Laevigata)
=======
<img width="778" alt="1739 - language display in new etd" src="https://user-images.githubusercontent.com/3064318/44472293-75652900-a5f3-11e8-9ec0-9898196bc9cf.png">

SHOW VIEW (Migrated Legacy ETD)
=======
<img width="756" alt="1739 - language display in migrated etd" src="https://user-images.githubusercontent.com/3064318/44472308-7a29dd00-a5f3-11e8-9c0c-68da7d0175e1.png">

EDIT VIEW (Migrated Legacy ETD)
========
<img width="1165" alt="1739 - language display in migrated etd - edit view" src="https://user-images.githubusercontent.com/3064318/44472322-82821800-a5f3-11e8-85a8-96535abe3b32.png">

mark-dce:
FYI - I am updating all records on staging that lack a language entry to have `language = ["English"]` for acceptance testing.  

We will need to update the migration code read the <mods:language> node from the source.  (see http://repo.library.emory.edu/fedora/objects/emory:cqt78/datastreams/MODS/content for an example of a submission in French)

mark-dce:
Test batch confirmed and checked on upgrade-etd this morning (migration against v2.0.0)

PR #1745, opened 2018-08-23
Add language to migration mapper

no-reply:
Map language when creating migrated resources. Default to "English" in the event no language is given.


mark-dce:
connects to #1739

PR #1754, opened 2018-08-24
Fixed a problem where when you delete all the partnering_agencys from the form

val99erie:
the old values are cached in the InProgressEtd record, so when
you reload the form, the old values still show up.

If you submit blank strings from the front end, you can effectively
delete the old values.


val99erie:
Don't merge.  I just found a problem.

val99erie:
ok, should be fixed now

Issue #1760, opened 2018-08-27
Terms of Use are out of date


mark-dce:
**ISSUE**
Release candidates servers are not displaying the proper "Terms of Use" 

**DETAILS**
The proper terms of use text is available at https://github.com/curationexperts/laevigata/blob/master/app/views/hyrax/static/terms.html.erb.  This is not the text being displayed on our testing servers.

The proper text is being displayed on the production system: 
https://etd.library.emory.edu/terms?locale=en  
I can't even figure out where the text being displayed on the qa server is coming from
https://qa-upgrade-etd.emory.edu/terms?locale=en <-- i.e. I can't even find this text if I search the code repo

**ACCEPTANCE**
- [x] When I click on the the Terms of Use link on the homepage, I am provided the proper terms of use

To Access Terms of Use
=============
<img width="719" alt="1760 access terms of use" src="https://user-images.githubusercontent.com/3064318/44672771-0e7ab200-a9ef-11e8-972b-912fd051b736.png">

Properly Updated Terms
===============
<img width="1439" alt="1760 correct terms of use" src="https://user-images.githubusercontent.com/3064318/44672772-0e7ab200-a9ef-11e8-9c23-a22348830964.png">

Obsolete Terms being displayed in release testing
==============
<img width="1438" alt="1760 obsolete terms of use" src="https://user-images.githubusercontent.com/3064318/44672773-0e7ab200-a9ef-11e8-931d-98769b24cd46.png">

mark-dce:
👍

PR #1761, opened 2018-08-27
Re-generate Terms of Use Content Block

bess:
Reload the Terms of Use Content Block from the file
in version control.

Run it like this:
```
RAILS_ENV=production bundle exec rake emory:terms_of_service
```
See https://staging-upgrade-etd.emory.edu/terms?locale=en

Connected to #1760


bess:
@mark-dce Fixed. I agree that's better, except that we need to update the `terms_page`, not the `agreement_page`.

mark-dce:
I'm merging despite the travis error - which appears to be due to a flappy test which I've written up here #1762 Intermittent CI Failure - EtdEmbargoService.assets_under_embargo

PR #1818, opened 2018-10-03
Make a csv report for admins more generic, fix committee members bug

alexBLR:
Add attributes to the rake command to generate a report for any school, academic year, program
Fix a bug when only one committee member and a chair is displayed in the report.


alexBLR:
@bess, I will go back and provide more documentation for the rake task.

PR #1820, opened 2018-10-22
Changes default ordering of user notifications

devanshu-m:
This commit orders user notifications by date (descending) to show most recent first.

Closes emory-libraries/etd-issues#57

### DIFF
--- a/app/views/hyrax/notifications/_notifications.html.erb
+++ b/app/views/hyrax/notifications/_notifications.html.erb
@@ -1,0 +1,39 @@
+<% if messages.present? %>
+  <div class="table-responsive">
+    <%# This file is copied from Hyrax. Notifications are ordered in ascending order of date by default. `data-order` attribute is used to change default order of date column to descending (most recent first) %>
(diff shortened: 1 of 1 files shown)


rotated8:
Please add a note in your commit message to say that this was copied in from Hyrax, and a comment in the file around where you changed it versus Hyrax.

app/views/hyrax/notifications/_notifications.html.erb
```diff
@@ -0,0 +1,38 @@
+<% if messages.present? %>
+  <div class="table-responsive">
+    <table class="table table-striped datatable" data-order='[[0,"desc"]]'>
```
rotated8: Please add a comment here about what's different between this file and the way it is in Hyrax.

devanshu-m: @rotated8 can you review now?

rotated8: 👍

Issue #1882, opened 2019-04-09
Upgrade webpack


RachelDCE:
Laevigata uses the latest Webpack

PR #1900, opened 2019-05-28
Updated Webpack configuration

little9:
This updates the webpacker gem, and the asscociated
webpack and babel packages to allow usage of the
newest webpack-dev-server. This also upgrades the yarn
packages and updates tests for any changes.

Connected to #1882


little9:
@bess I was able to deploy this to QA and did some manual testing:
 
https://qa-etd.library.emory.edu/concern/etds/3r074w01x?locale=en

We should be good to go on this

Issue #1903, opened 2019-06-04
Address low-risk omniauth security issue


mark-dce:
Follow the breadcrumbs here https://github.com/curationexperts/laevigata/network/alert/Gemfile.lock/omniauth/open  

and here 
https://github.com/omniauth/omniauth/pull/809
https://github.com/omniauth/omniauth-rails/pull/1

Determine whether there is any action needed on our part.

bess:
I took a swing at the omniauth security issue, but I don't think it's going to be a quick fix. See PR here: https://github.com/curationexperts/laevigata/pull/1907. 

This is not a new issue. It has been around since 2015, but a CVE was just issued for it, which is why it's suddenly on our radar. There will eventually be a better fix in place. I don't think we're particularly vulnerable, as first someone would have to compromise Emory's shibboleth server to take advantage of this. I recommend we wait until there is a straightforward fix available.

mark-dce:
Exploiting this vulnerability would require someone to
1) breach emory's shibboleth system
2) have access to an active session still open in another users browser

We think the likelihood of that (and the corresponding risk if it every did happen) is sufficiently low that we're closing out this issue for now.

PR #1907, opened 2019-06-04
DO NOT MERGE Address security issue CVE-2015-9284

bess:
Vulnerable versions: <= 1.9.0
Patched version: No fix
The request phase of the OmniAuth Ruby gem is vulnerable to Cross-Site Request Forgery when used as part of the Ruby on Rails framework, allowing accounts to be connected without user intent, user interaction, or feedback to the user. This permits a secondary account to be able to sign into the web application as the primary account.

More info available at https://nvd.nist.gov/vuln/detail/CVE-2015-9284

After review, laevigata does not appear to be particularly vulnerable to
this issue, as it would require that the Emory Shibboleth system also be
compromised. However, to be on the safe side, we are implementing the
recommended fix. Until the omniauth gem itself can be upgraded (long
thread about why this is taking so long available here:
https://github.com/omniauth/omniauth/pull/809) the recommendation is to
install the omniauth-rails_csrf_protection gem and ensure all
authentication requests happen via POST.

This PR installs the required gem, and all of our authentication
requests already use a POST protocol.


bess:
After deploying this branch to QA I found it doesn't work. When I try to log in via shibboleth I get this error:

login_broken

dankozlowski:
@bess Are you able to cont. work here, or are you calling this blocked?

bess:
Hi, @dankozlowski . I guess I'm feeling blocked at this point. My plan was to put this down for awhile until there was a more straightforward fix. Unless you've got a hint for me around what I'm doing wrong? Thanks!

mark-dce:
We've evaluated the risk and feel it is sufficiently low that we shouldn't expend more effort in this PR.

See #1903 for further details

Issue #1939, opened 2019-07-16
Wrong Timestamp on Review Submissions Tab


mark-dce:
**ISSUE**
https://curationexperts.freshdesk.com/a/tickets/107
> This can be an issue if an ETD is submitted close to midnight, where a reviewer may see an incorrect submission date on the Review Submissions tab. The correct date shows up elsewhere in the app. My uninformed guess is this date is based on UTC rather than EST time.

**ACCEPTANCE**
- [ ] Ensure that submission dates are displayed in relation to EST time on the "Review Submission" tab of the application.

PR #1942, opened 2019-07-17
Format date in workflow admin page

little9:
This adds a helper to format the
date in the workflow admin page so that
is displays the time in EST and using AM/PM.

Connected to #1939


little9:
@bess Thanks for catching that! It's in EST now.

PR #1994, opened 2020-02-14
DO NOT MERGE Approvers should always have edit rights

bess:
Alter the Laney workflow so that people who are approvers will always
have edit rights on the object.

DO NOT MERGE until we've heard back from Emory staff that this fixes the problem they reported.

Connected to https://github.com/curationexperts/in-house/issues/575


bess:
Consensus on Issue #125 is that approvers shouldn’t have the ability to change an object and that the student should go in and change it.

PR #2037, opened 2020-07-24
Check for workflow state before running graduation job

bess:
We never want to run the graduation job for a work that hasn't been approved.
Theoretically, that shouldn't ever happen, but since we do sometimes run this
job manually, better to check explicitly.


mark-dce:
Closing this PR because we've addressed some of the work in other PRs.  I've added ticket #2143 *GraduationJob should only publish "approved" ETDs* to cover the remaining work.

PR #2067, opened 2020-11-08
Visibility fixes

mark-dce:
Add tests and code to handle visibility edge cases that represent inconsistent state 
between embargo length and embargo flags.


mark-dce:
Superceeded by new PR

Issue #2065, opened 2020-10-24
Embargo visibility is not set correctly on newly created and submitted ETDs


mark-dce:
**ISSUE**
When I submit a new ETD with no embargo request, it appears with a "Restricted; Files Only" badge.

**ACCEPTANCE**
- [ ] ETDs submitted with no embargo request should have an "Open" badge

<img width="468" alt="image" src="https://user-images.githubusercontent.com/3064318/97094942-502f3380-161f-11eb-96af-bc61ed59ea62.png">

<img width="236" alt="image" src="https://user-images.githubusercontent.com/3064318/97094955-689f4e00-161f-11eb-806f-4612b60e2085.png">

mark-dce:
The root cause appears to be a flaw in the VisibiityTranslator that does not handle missing `embargo_length` values correctly.

At present, there are 295 published ETDs in production that appear to have this issue.  The following solr query will return the current list on any environment:
http://localhost:8990/solr/laevigata/select?fl=id,%20files_embargoed_bsi,%20visibility_ssi,%20workflow_state_name_ssim,%20title_tesim&indent=on&q=has_model_ssim:Etd%20AND%20visibility_ssi:files_restricted%20AND%20files_embargoed_bsi:false%20AND%20workflow_state_name_ssim:published&rows=400&sort=date_modified_dtsi%20DESC&wt=csv

ETDs should not be able to enter a state where visibility is `files_resticted` but `files_embargoed` is false

mark-dce:
The fixes in [PR#2089](https://github.com/curationexperts/laevigata/pull/2089) address the underlying cause of this issue.  Once the PR has been merged, released, and deployed, impacted ETDs will need to be re-indexed to show the correct badges.

PR #2089, opened 2021-02-24
Update VisibilityTranslator to handle various edge cases

mark-dce:
Addresses the issue described in #2065 

* Treat `embargo_length == nil` the same as `embargo_length == "None"`
* Restrict works with conflicting embargo settings - i.e. `embargo_length == "6 months"`
   BUT `files_embargoed == toc_embargoed == abstract_embargoed == false`
* Add logger warnings and errors to trace unexpected states

**BEFORE**
image
image

**AFTER**
image
image

### DIFF
--- a/app/services/visibility_translator.rb
+++ b/app/services/visibility_translator.rb
@@ -17,6 +17,7 @@
 #
 # @see Hydra::AccessControls::Visibility
 class VisibilityTranslator
+  RESTRICTED      = Hydra::AccessControls::AccessRight::VISIBILITY_TEXT_VALUE_PRIVATE
   ALL_EMBARGOED   = 'all_restricted'.freeze
   FILES_EMBARGOED = 'files_restricted'.freeze
   TOC_EMBARGOED   = 'toc_restricted'.freeze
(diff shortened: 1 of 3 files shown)


fnibbit (approved):

app/services/visibility_translator.rb
```diff
@@ -44,10 +45,14 @@ def self.visibility(obj:)
   # When checking whether a value is true, check whether it has a "true" string too.
   def visibility
     return proxy.visibility if obj.hidden? || !obj.under_embargo?
+
+    return OPEN             if embargo_length_none?(obj)
     return ALL_EMBARGOED    if obj.abstract_embargoed.to_s == "true"
```
maxkadel, line 50: I wonder why this has to be forced to a string and compared to a string, rather than being a boolean? Don't know if it's relevant to the weirdness we're having, but it makes me curious.

mark-dce: There's legacy data migrated that has string values while the new application uses booleans, so the field might have either `"true"` or '`true` stored in Fedora.   

I would have preferred to normalize the data at the time of migration, but that's not the choice the team made at the time.

I think we could consider doing that normalization now - I think it would make the data and the code simpler and easier to understand.

PR #2092, opened 2021-02-25
Update indexing

mark-dce:
bab9817
All of the attributes here are defined in both
**/app/models/schemas/emory_etd_schema.rb**
and
**/app/models/etd.rb**

This is the first step to defining all attributes in a single location.

This PR also resolves this type of warning message in logs:
```
W, [2021-02-24T20:01:54.009886 #962]  WARN -- : Same predicate (http://purl.org/spar/pso/embargoed#abstract) used for properties abstract_embargoed and abstract_embargoed
```


mark-dce:
Cleaner version in new PR

PR #2097, opened 2021-02-25
Remove scheduled runs of graduation rake task

fnibbit:


mark-dce (approved):
Looks good.  

And looking at it now, I suspect this might have required `bundle exec` to run correctly anyway.

PR #2106, opened 2021-03-02
Add simplecov for formatting and ignoring files for test coverage

maxkadel:


bin/rails
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env ruby
+if ENV['RAILS_ENV'] == 'test'
+  require 'simplecov'
+  SimpleCov.start 'rails'
```
mark-dce, line 4: I don't know that we're always invoking things via binstubs - would it make sense to move this to the rails helper - i.e. see the pattern here: https://github.com/RepoCamp/ascteam2/commit/8eeb14a6abce68ea807818ca5a4dabaf2929a9d2

maxkadel: Aaaah, ok, I was over-generalizing from step 2 here but I think you're right - https://github.com/simplecov-ruby/simplecov#getting-started

fnibbit (approved):

PR #2124, opened 2021-03-14
Refactor workflow setup to improve performance

mark-dce:
This change moves a relatively expensive operation that loads
workflow configuration from the inside of a loop that gets executed
many times and runs it only once afterwards.

Invoking load_workflows reloads workflows for all existing admin_sets.
If we do this after each new admin set, the first call runs for 1 set,
the second call for 2 sets, the third for 3 sets, etc. resulting in
1+2+3+...+n reloads.  This exponential performance gives us a cost of
n(n+1)/2 for the whole process.

Since we're not running any other operations in between, we can just
reload all the workflows after we've created all the admin sets for
a linear cost of n (the number of admin sets) for the whole process.

The intent of this change is to improve performance without changing
behavior, so the test suite should run (faster) without needing to
be modified.

**BEFORE**
Test suite run statistics on Mark's local system
```
Finished in 62 minutes 59 seconds (files took 18.18 seconds to load)
710 examples, 0 failures, 9 pending

Randomized with seed 41307
```

**AFTER**
```
Finished in 30 minutes 34 seconds (files took 17.01 seconds to load)
710 examples, 0 failures, 9 pending

Randomized with seed 4098
```

### DIFF
--- a/lib/workflow_setup.rb
+++ b/lib/workflow_setup.rb
@@ -40,6 +40,7 @@ class WorkflowSetup
       @logger.debug "Attempting to make admin set for #{as}"
       make_admin_set_from_config(as)
     end
+    load_workflows
     everyone_can_deposit_everywhere
     give_superusers_superpowers
   end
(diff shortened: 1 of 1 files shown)


mark-dce:
Capistrano Deployment

**BEFORE ~7 minutes**
```
06:56 deploy:log_revision
      01 echo "Branch main (at 7b65ef2d51e9c8f1c556e4c96f6f6b3046905bb7) deployed as release 20210315023628 by mark" >…
    ✔ 01 deploy@qa-etd.library.emory.edu 0.103s
```

**AFTER ~ 4 minutes**
```
03:59 deploy:log_revision
      01 echo "Branch debug-qa (at a2f4ebda6c5e87c4301bd00ac29ebeccfcec9baf) deployed as release 20210315024627 by mar…
    ✔ 01 deploy@qa-etd.library.emory.edu 0.105s
```

mark-dce:
Closing and will re-issue to trigger CI

PR #2165, opened 2021-09-10
Upgrade to Hyrax 2.9.6 to fix first-deploy bug

maxkadel:


mlooney (approved):

maxkadel:
For context, see https://github.com/samvera/hyrax/pull/4718 

The first time deploying to a new environment was failing, because there was an order dependency for FlipFlop inclusion.

PR #2178, opened 2021-10-10
Flag slow test as a smoke_test

mark-dce:
This test doesn't test any code that isn't tested elsewhere,
it just tests it more times by runing the full production config file
that sets up 21 admin sets - i.e. it runs the same tests 21 times.

While it's somewhat useful to know that all currently configured
admin sets in config/emory/admin_sets.yml are setup correctly, that
can be manually verified in the dashboard on the page at
https://etd.library.emory.edu/schools

Long term, this check probably belongs in some kind of deployment check
rather than the main test suite.

### DIFF
--- a/spec/lib/workflow_setup_spec.rb
+++ b/spec/lib/workflow_setup_spec.rb
@@ -234,7 +234,7 @@ RSpec.describe WorkflowSetup, :clean do
       end
     end
   end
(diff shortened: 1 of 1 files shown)


mark-dce:
I'm merging this since coveralls reports no change in coverage and it speeds up the test suite.

PR #2209, opened 2022-03-02
Schoolie

mlooney:


lib/tasks/schoolie.rake
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "json"
+namespace :schoolie do
+  task sitemap: :environment do
+    ids = JSON.parse(`curl -s 'http://127.0.0.1:8983/solr/laevigata/select?fq=workflow_state_name_ssim:(published%20OR%20approved)&fl=id&rows=15000&sort=date_modified_dtsim%20ASC'`)["response"]["docs"].map do |x| # rubocop:disable Layout/LineLength
```
mark-dce, line 6: we should use a Hyrax/blacklight search service here.  But this is good for now.

mark-dce (approved):

PR #2224, opened 2022-03-18
Silences redis deprecations

mlooney:


mark-dce (changes requested):
Is there a way to resolve the deprecations instead of just silencing them.  I'm anxious about future upgrades if we just silence them.

It seemed like we needed to update sidekiq and remove some setting from our redis config file.

mark-dce:
Sounds like Hyrax isn't ready for the permanent fix, so we'll go with silencing the noise for now.

mark-dce (approved):

PR #2226, opened 2022-03-22
Copies the deactivate functionality from tenejo

norm0:


mark-dce:
@norm0 - there ended up being more back-end code that needed to be pulled over, so I ended up making https://github.com/curationexperts/laevigata/pull/2232

PR #2232, opened 2022-03-28
Allow users to be deactivated and reactivated

mark-dce:
We want to be able to prevent administrative users from logging in
if their roles change without nececessarily deleting their activity
history - i.e. files they may have uploaded to student ETDs and/or
approvals or change requests they have issued.

This change adds the user interface and back-end code to allow
users to be deactivated (or soft-deleted) so that they can no
longer access the system, while any records tied to their userID
still point to a valid (but disabled for login) User record.


mlooney (approved):

mark-dce:
image


image

PR #2234, opened 2022-04-05
Nightly reports

mlooney:
- Adds report job
- properly handles activetriples'
- Finds only graduated etds, handles multivalue fields
- Adds uid to report, and moves report to rake task
- Addds private file controller & localized task links to admin
- Added request spec


config/locales/en.yml
```diff
@@ -20,6 +20,7 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
+  download_report: "Download Report"
```
mark-dce, line 23: Per Rachel - can we change this to "Download Graduation Report"

mark-dce (changes requested):
Looks good, but can we change the link name before merging?

lib/tasks/nightly_report.rake
```diff
@@ -0,0 +1,37 @@
...
+            when "uid"
+              User.find_by(ppid: etd.depositor).uid
+            when "graduation_year"
+              etd.graduation_date
+            when "abstract_length"
+              etd.abstract.first.length
+            else
+              k = etd.send(x)
+              if k.is_a?(ActiveTriples::Relation)
+                k.map { k.first }.join('; ').tr("\n", ' ')
```
mark-dce, line 23: Need to update this to handle multiple values correctly - i.e. send an index instead of .first

mark-dce (changes requested):

mark-dce (approved):

PR #2235, opened 2022-04-05
Userup

mlooney:


lib/tasks/sync_user.rake
```diff
@@ -0,0 +1,21 @@
+require 'yaml'
+
+namespace :emory do
+  task sync_users: :environment do
```
mark-dce, line 4: maybe call the task update_approvers so we call   
`rake emory:update_approvers`. 
instead of   
`rake emory:sync_users`. <-- sync which users with what?

mark-dce (changes requested):

mark-dce (approved):

PR #2265, opened 2022-05-01
Proquest Preliminary Refactor

mark-dce:
* Handle timezones for graduation dates. 
    * Consistently use a Time class object for degree_awarded
    * Handle conversion to application time-zone as needed
    * Update specs appropriately
* Move the .submit_to_proquest? logic 
    * Move the business logic out of the Job class and into the ProquestBehaviors module
    * Refactor the rules slightly for clarity
* Get the ProquestJob spec working
    * Add necessary setup to allow spec to run
    * Replace instance variables in the ProquestJob with method call arguments
    * Inline the ProquestCompletionJob - a separate job provided no added functionality
* Merge graduation_service_spec.rb and graduation_service_method_spec.rb
    * Use contexts to limit expensive test setup (instead of splitting tests into two files)


spec/factories/etd.rb
```diff
@@ -179,6 +179,11 @@
         degree_awarded { (Time.zone.today - 1.week).strftime('%Y-%m-%d') }
         submitting_type { [] << "Dissertation" }
         school { ["Laney Graduate School"] }
+        depositor do
+          u = User.new(uid: FFaker::Internet.user_name, ppid: 'P0000005', display_name: creator.first)
```
mlooney, line 183: use create() or create!(), or check the return value from save(), since it might fail and continue on like nothing happened

mlooney (changes requested):

mlooney (approved):

PR #2332, opened 2023-08-17
Update Java patch version

mark-dce:


mark-dce:
Figured out it was a Solr version issue in CI breaking builds.  See #2334 2334

PR #2333, opened 2023-08-17
Trival change to test CI

mark-dce:


mark-dce:
Figured out it was a Solr version issue in CI breaking builds.  See #2334

PR #2415, opened 2024-09-20
[REFACTOR] Remove unnecessary form code

mark-dce:
**ISSUE**
We previously provided code pathways for native Hyrax conroller updates if we were not using the JavaScript front-end, and separate processing for patches from the JavaScript front-end.

All user input now comes from the JavaScript front-end, so we no longer need to support multiple update pathways.

**RESOLUTION**
Therefore, we can remove the `request_from_form` flag that triggers the JavaScript update pathway - i.e. all updates should now use that single update process.


mark-dce:
Embargo management uses the ETD controller using rails forms, so this code turns out not to be obsolete.

PR #2501, opened 2026-04-22
Add editng support for inactive departments and graduation semesters

mark-dce:
This change adds two features:
1. Approvers and SuperAdmins to choose inactive departments and graduation semesters. Inactive options are flagged with a special icon.
2. Allow submitters to save and re-submit InProgress ETDs with an inactive department or graduation semester.

### DIFF
--- a/app/javascript/Department.vue
+++ b/app/javascript/Department.vue
@@ -1,67 +1,59 @@
 <template>
-  <div>
-    <!-- TODO: adding v-model='selected' enables the selected value to appear in the case of an ipe_etd, but it will not without it, with a hyrax etd. But we should refactor for a better solution. -->
-    <label for="department"> {{ sharedState.getDepartmentHeading() }} </label>
-    <select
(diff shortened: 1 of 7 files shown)


app/javascript/Department.vue
```diff
@@ -1,67 +1,59 @@
...
+    this.selected = this.sharedState.getSavedDepartment()
+    this.fetchData()
+    this.sharedState.getSubfields()
   },
-  mounted: function() {
-    this.$nextTick(function() {
-      //this is to handle the case of a saved department
-      if (_.has(this.sharedState.savedData, "etd_id")) {
-        var selectedArray = this.sharedState.getSavedOrSelectedDepartment();
-        this.selected = selectedArray[0];
```
mark-dce, line 51: Removing this introduced a bug when editing an existing ETD because Rails was sending an array instead of a simple value.

FIxed in https://github.com/curationexperts/laevigata/pull/2503

